PR #3, opened 2022-12-02
Add initial tox/black/mypy config

naglepuff:
This PR adds some testing for Image DePHI, which can be built upon to accommodate more kinds of testing.

We are using `tox` to configure tests. For now, linting, import sorting, type checking, and formatting have been added.

### DIFF
--- a/imagedephi.py
+++ b/imagedephi.py
@@ -2,7 +2,6 @@ import argparse
 from enum import Enum
 from pathlib import Path
 import sys
-from typing import Dict, List
 
 import tifftools
 from tifftools import Datatype, TiffTag
@@ -13,7 +12,7 @@ class RedactMethod(Enum):
     DELETE = 2
 
 
-def get_tags_to_redact() -> Dict[int, Dict]:
+def get_tags_to_redact() -> dict[int, dict]:
     return {
         270: {
             'id': 270,
@@ -24,14 +23,14 @@ def get_tags_to_redact() -> Dict[int, Dict]:
     }
 
 
-def redact_one_tag(ifd: Dict, tag: TiffTag, redact_instructions: Dict) -> None:
+def redact_one_tag(ifd: dict, tag: TiffTag, redact_instructions: dict) -> None:
     if redact_instructions['method'] == RedactMethod.REPLACE:
         ifd['tags'][tag.value]['data'] = redact_instructions['replace_value']
     elif redact_instructions['method'] == RedactMethod.DELETE:
         del ifd['tags'][tag.value]
 
 
-def redact_tiff_tags(ifds: List[Dict], tags_to_redact: Dict[int, Dict]) -> None:
+def redact_tiff_tags(ifds: list[dict], tags_to_redact: dict[int, dict]) -> None:
     for ifd in ifds:
         sub_ifd_list = []
         for tag, tag_info in sorted(ifd['tags'].items()):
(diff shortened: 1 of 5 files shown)


imagedephi.py
```diff
@@ -1,36 +1,37 @@
...
+from tifftools import Datatype, TiffTag
 
 
 class RedactMethod(Enum):
     REPLACE = 1
     DELETE = 2
 
 
-def get_tags_to_redact():
+def get_tags_to_redact() -> Dict[int, Dict]:
```
brianhelba, line 16: FYI, eventually it might help to turn this tag specification into a Data Class, to make is more structured than just a dict, but I think that's out of scope of this PR.

requirements.txt
```diff
@@ -1,3 +1,6 @@
 black
```
brianhelba, line 1: I'm not sure if we need packages like `black` to be installed directly as dependencies. Tox will install it for us into a dedicated virtual env.

However, this whole requirements format will need to be revisited once we move things into a package, so I don't think it matters too much now.

imagedephi.py
```diff
@@ -47,21 +48,21 @@ def redact_tiff_tags(ifds, tags_to_redact):
...
-    return output_dir/f'REDACTED_{file_path.name}'
+def get_output_path(file_path: Path, output_dir: Path) -> Path:
+    return output_dir / f'REDACTED_{file_path.name}'
 
 
-def redact_images(image_dir, output_dir):
+def redact_images(image_dir: Path, output_dir: Path) -> None:
     for child in image_dir.iterdir():
         try:
             tiff_info = tifftools.read_tiff(child)
```
brianhelba, line 68: In the future, I'd like to see if we can add stronger annotations for the `tifftools` API as local stubs. I can look into this once this PR is merged.

brianhelba (changes requested):

manthey:
I'll just note that we don't want to go too far down the route of tifftools as the reader/writer, since it will only work on tiff files.  We will want to read a much broader range of files and write them back out in their original format if possible and a selected format if not.  Further, tifftools can read ndpi files but will write TIFF files, not ndpi (since ndpi is a bizarre misuse of some of tiff).

imagedephi.py
naglepuff: I opened #6 to track this.

tox.ini
```diff
@@ -0,0 +1,49 @@
...
+deps =
+    black
+    isort
+commands =
+    isort {posargs:.}
+    black {posargs:.}
+
+[testenv:type]
+skipsdist = true
+skip_install = true
```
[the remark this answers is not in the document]
naglepuff: We had a chat about this IRL. Removing these lines doesn't work for us right now because we don't have a package to install. If `tifftools` had type annotations we could add it under `deps`, but for now we can leave this as is.

imagedephi.py
```diff
@@ -1,36 +1,37 @@
 import argparse
-import sys
 from enum import Enum
 from pathlib import Path
+import sys
+from typing import Dict, List
```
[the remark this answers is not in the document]
naglepuff: Thanks. 1ef11d7 contains this change

mypy.ini
```diff
@@ -0,0 +1,3 @@
+[mypy]
```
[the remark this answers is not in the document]
naglepuff: Moved in 38eb32a

tox.ini
```diff
@@ -0,0 +1,49 @@
...
+[testenv:lint]
+skipsdist = true
+skip_install = true
+deps =
+    flake8<6
+    flake8-black
+    flake8-bugbear
+    flake8-docstrings
+    flake8-isort
+    flake8-quotes
```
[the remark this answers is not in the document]
naglepuff: Nope, must have missed that plugin. Added in a8ddf59

requirements.txt
naglepuff: Cleaned up `requirements.txt` in e5d9f71

brianhelba (approved):

PR #8, opened 2022-12-09
Improve type annotations

### LINKED ISSUE #6: Investigate stronger type annotations for `tifftools` with stubs

brianhelba:
Fixes #6.


imagedephi.py
```diff
@@ -23,34 +27,31 @@ def get_tags_to_redact() -> dict[int, dict]:
...
     if redact_instructions['method'] == RedactMethod.REPLACE:
         ifd['tags'][tag.value]['data'] = redact_instructions['replace_value']
     elif redact_instructions['method'] == RedactMethod.DELETE:
         del ifd['tags'][tag.value]
 
 
-def redact_tiff_tags(ifds: list[dict], tags_to_redact: dict[int, dict]) -> None:
+def redact_tiff_tags(ifds: list[IFD], tags_to_redact: dict[int, dict[str, Any]]) -> None:
     for ifd in ifds:
-        sub_ifd_list = []
```
brianhelba, line 35: Note, this variable was never being consumed, so I just removed it.

imagedephi.py
```diff
@@ -23,34 +27,31 @@ def get_tags_to_redact() -> dict[int, dict]:
...
-        for tag, tag_info in sorted(ifd['tags'].items()):
-            tag = tifftools.commands.get_or_create_tag(
-                tag,
-                tifftools.Tag,
-                {'datatype': Datatype[tag_info['datatype']]},
```
brianhelba, line 40: This should have been a keyword argument, rather than a dict.

imagedephi.py
```diff
@@ -23,34 +27,31 @@ def get_tags_to_redact() -> dict[int, dict]:
...
     elif redact_instructions['method'] == RedactMethod.DELETE:
         del ifd['tags'][tag.value]
 
 
-def redact_tiff_tags(ifds: list[dict], tags_to_redact: dict[int, dict]) -> None:
+def redact_tiff_tags(ifds: list[IFD], tags_to_redact: dict[int, dict[str, Any]]) -> None:
     for ifd in ifds:
-        sub_ifd_list = []
-        for tag, tag_info in sorted(ifd['tags'].items()):
-            tag = tifftools.commands.get_or_create_tag(
```
brianhelba, line 37: This actually lived under `tifftools.constants`.

stubs/tifftools/__init__.pyi
```diff
@@ -0,0 +1,18 @@
+from .constants import Datatype, Tag, TiffDatatype, TiffTag
+from .exceptions import MustBeBigTiffError, TifftoolsError, UnknownTagError
+from .tifftools import read_tiff, write_tiff
+
+__version__: str
+
+__all__ = [
```
brianhelba, line 7: This is certainly not complete, but it at least covers all the parts of the API that we're currently using.

pyproject.toml
```diff
@@ -12,3 +12,7 @@ combine_as_imports = true
 [tool.mypy]
 ignore_missing_imports = true
 show_error_codes = true
+mypy_path = "$MYPY_CONFIG_FILE_DIR/stubs"
+
+[tool.pyright]
```
brianhelba, line 17: This will make VSCode nicely pick up the stubs without further configuration. Maybe we want to go all-in on VSCode as a preferred tool for developers, but that's a bigger commitment and separate discussion.

stubs/tifftools/tifftools.pyi
```diff
@@ -0,0 +1,41 @@
...
+    size: int
+    bigEndian: bool
+    bigtiff: bool
+    tagcount: int
+
+class TagEntry(TypedDict):
+    datatype: int
+    count: int
+    datapos: int
+    offset: NotRequired[int]
```
brianhelba, line 29: Having an optional item in a `TypedDict` requires Python 3.11 to properly annotate: https://docs.python.org/3/whatsnew/3.11.html#pep-655-marking-individual-typeddict-items-as-required-or-not-required

This forces us to bump to Python 3.11 for the whole project, but I think having correct stubs is a strong enough reason to do so. I expected that we'd likely upgrade at some point regardless.

brianhelba:
@naglepuff PTAL. I've left comments for the most noteworthy (to me) things.

stubs/tifftools/tifftools.pyi
manthey: I don't think requiring Python 3.11 is a good idea *yet*.  Some of our tile sources in large_image don't yet have wheels published for 3.11 (e.g., some of the dependencies for nd2 files and the tifffile library).  And, our collaborators probably haven't upgraded to 3.11 anywhere yet, either, so it will add unnecessary complication to their use of it.

naglepuff: What about using [totality](https://mypy.readthedocs.io/en/latest/more_types.html#totality)? Maybe this doesn't solve the problem completely but it seems like a compromise between staying at 3.10 and having helpful type annotations.

stubs/tifftools/constants.pyi
```diff
@@ -0,0 +1,47 @@
+from typing import Any, Generator, Generic, TypeVar, overload
+
+_ConstantAttr = Any
```
naglepuff, line 3: What is this doing? Just aliasing `Any`?

stubs/tifftools/constants.pyi
```diff
@@ -0,0 +1,47 @@
...
+
+class TiffConstant(int):
+    value: int
+    name: str
+    def __init__(self, value: int, constantDict: dict[str, _ConstantAttr]) -> None: ...
+    def __getitem__(self, key: str) -> _ConstantAttr: ...
+    def get(self, key: str, default: _ConstantAttr = ...) -> _ConstantAttr: ...
+
+_ConstantT = TypeVar('_ConstantT', bound=TiffConstant)
```
naglepuff, line 12: My understanding is that this line creates a new type called `_ConstantT` which represents `TiffConstant` and any subtype of `TiffConstant`. Is that correct?

In that case, I think the name `_TiffConstantT` is more clear.

naglepuff (changes requested):
For the most part, these changes are straightforward and I like them. Having type annotations will make development easier. I had a couple of questions about the subs.

I'm hesitant to mark this as approved, since @manthey brings up some good points about switching to 3.11 too quickly. If there's a sensible workaround that can be done here I think it makes sense to do that to keep these PRs moving, and an issue can capture the shortcomings of a workaround. Otherwise I feel this change requires more discussion before merging.

stubs/tifftools/tifftools.pyi
brianhelba: See #16. I'll change this PR now to use `total`.

stubs/tifftools/constants.pyi
brianhelba: Agreed. I'll change the name.

stubs/tifftools/constants.pyi
brianhelba: Yeah, I was trying use this alias to conceptually separate the type for `TiffConstant` attribute values, indicating all the related places that it's used. I can add a code comment if you think this is unclear.

brianhelba: I changed the name, added an explicit `TypeAlias` type, and added a comment.

brianhelba:
@naglepuff PTAL again. I've addressed your feedback.

If it helps to simplify things, feel free to resolve any of the line comments I added, once you've reviewed / accept them.

PR #10, opened 2022-12-09
Allow Black to handle string quote formatting

brianhelba:
This has several benefits:
* String quotes are now auto-formatted correctly; no need to fix quote-related Flake8 errors by hand.
  * Continue to type the quotes however you want. Black will just fix them afterwards.
* We can unpin Flake8.

Note the addition of `.git-blame-ignore-revs`, to avoid `git blame` attribution of these automated changes. See https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/ and [run `git config --global blame.ignoreRevsFile .git-blame-ignore-revs` locally if you want automatic support](https://michaelheap.com/git-ignore-rev/).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Reformat code to standardize double quoted strings
+bb0159f16a038c8689c242c575607c31a843aed7
--- a/imagedephi.py
+++ b/imagedephi.py
@@ -18,27 +18,27 @@ class RedactMethod(Enum):
 def get_tags_to_redact() -> dict[int, dict[str, Any]]:
     return {
         270: {
-            'id': 270,
-            'name': 'ImageDescription',
-            'method': RedactMethod.REPLACE,
-            'replace_value': 'Redacted by ImageDePHI',
+            "id": 270,
+            "name": "ImageDescription",
+            "method": RedactMethod.REPLACE,
+            "replace_value": "Redacted by ImageDePHI",
         }
     }
 
 
 def redact_one_tag(ifd: IFD, tag: tifftools.TiffTag, redact_instructions: dict[str, Any]) -> None:
-    if redact_instructions['method'] == RedactMethod.REPLACE:
-        ifd['tags'][tag.value]['data'] = redact_instructions['replace_value']
-    elif redact_instructions['method'] == RedactMethod.DELETE:
-        del ifd['tags'][tag.value]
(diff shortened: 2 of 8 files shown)


naglepuff (changes requested):
I like the idea of making developers' lives easier. Can we make this auto-format to single quotes? Other projects in the Digital Slide Archive I've worked on all tend to use single quotes.

Leengit:
There are no doubt lots of opinions on this.  I am in favor of using all `black` defaults except the command line `--skip-magic-trailing-comma` (aka `-C`).  This promotes all Python code having the same look and feel even outside DigitalSlideArchive.  In the name of sacrificing for uniformity, this means a line length of 88, which I would otherwise consider too short.

Details: the `-C` tells `black` to treat
```python
a = [b, c]
```
and
```python
a = [b,
     c,]
```
the same, which is to put it all on one line if possible within 88 characters; otherwise one  list element per line.  Without `-C` the behavior is to choose one line if there is no comma after the last element (and it fits into 88 characters), and to choose one list element per line if there is a comma there.  (In particular, using `-C` doesn't deviate from the look of other repositories that don't use it because a call to `black` after `black -C` won't change anything.)

jcfr:
For new project, I suggest to standardize on double quotes and maximize use of existing tooling [^3].

[^3]: https://github.com/scikit-build/scikit-build/blob/359bb6ce59506b6527121fbea5ff3e4c8eb05396/.pre-commit-config.yaml#L35-L46

The use of `.git-blame-ignore-revs` [^1] helps to ensure `blame` works as expected both locally and with the GitHub UI.

To understand the rational behind the standardization, you may also be interested in discussion we had in the context of scikit-build [^2]

[^1]: https://github.com/Slicer/Slicer/pull/6306
[^2]: https://github.com/scikit-build/scikit-build/pull/665#issuecomment-1039573165

brianhelba:
Black is intentionally not configurable to use single quotes. See: https://github.com/psf/black/issues/118#issuecomment-380171356

The only alternate is the status quo, where we don't use Black to format quotes and force developers to fix things by hand.

manthey:
Black isn't the only auto-formatter.  autopep8 has been around for a decade, whereas black only came out of beta a year ago.

naglepuff (approved):
Sure, let's go for it. I'm fine following black's decision, so we can move on to more useful development.

PR #11, opened 2022-12-09
Add an editorconfig file

brianhelba:
This will ensure that our editors know how many spaces to use for indentation.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,28 @@
+root = true
+
+[*]
+indent_style = space
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,28 @@
...
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+[*.py]
+indent_size = 4
+max_line_length = 100
+
+[*.js]
```
brianhelba, line 14: We don't have `.js`, `.html`, or `.css` files yet. It's foreseeable that the GUI will require some of these formats, so I'm proactively adding them.

PR #12, opened 2022-12-09
Upgrade to Tox 4 and use editable installs

brianhelba:
This makes `tox` significantly faster to run.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,15 @@
 [tox]
-envlist =
+# Don't use "min_version", to ensure Tox 3 respects this
+minversion = 4
+env_list =
     lint,
     type
 
+[testenv]
+base_python = python3.10
+use_develop = true
+
 [testenv:lint]
-skipsdist = true
 skip_install = true
 deps =
     flake8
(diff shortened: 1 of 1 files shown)


brianhelba:
Depends on some other PRs to be merged first. Also, I want to ensure that our requirements are picked up for the `type` environment, once we get a Python package setup.

tox.ini
```diff
@@ -1,10 +1,15 @@
...
-envlist =
+# Don't use "min_version", to ensure Tox 3 respects this
+minversion = 4
+env_list =
     lint,
     type
 
+[testenv]
+base_python = python3.10
+use_develop = true
```
brianhelba, line 10: FYI, it's generally safe to run tests against an editable install for an application like this.

If we were building a library where we needed to guarantee that the sdist is correct, we'd want to be installing the package into an isolated test environment, to ensure that `import imagedephi` worked properly.

brianhelba:
@naglepuff Let me know if this works for you.

naglepuff (approved):
LGTM

PR #13, opened 2022-12-09
Run type checks in strict mode

brianhelba:


brianhelba:
Let's get #8 merged first, but this hopefully will ensure that autocomplete is always maximally helpful.

PR #14, opened 2022-12-12
Add better argument validation

naglepuff:


naglepuff:
Now, the input and output paths are validated by [argparse](https://docs.python.org/3/library/argparse.html#type), not our main method.

PR #15, opened 2022-12-12
Add setup.py

naglepuff:


jcfr:
For new package, I suggest to provide a `pyproject.toml` [^1] instead of legacy `setup.py`.

[^1]: https://pip.pypa.io/en/stable/reference/build-system/

setup.py
```diff
@@ -0,0 +1,7 @@
+from setuptools import setup
+
+setup(
+    name='Image DePHI',
```
brianhelba, line 4:
```suggestion
    name='ImageDePHI',
```
This shouldn't have spaces, since it's used to build the package file name.

brianhelba (changes requested):
A change to the package name is needed.

brianhelba:
We can discuss `pyproject.toml` in #21. I think this PR is fine using the legacy mechanism.

brianhelba (approved):

PR #19, opened 2022-12-12
Create initial functional tests

### LINKED ISSUE #4: Add functional tests

naglepuff:
Resolves #4 

Now our testing suite contains a new environment called `test`, which will run `pytest` tests for Image DePHI. There is now a single test that tests the `redact_images` method, as well as some test data (a blank tiff image created by `large_image_converter`).

### DIFF
--- a/tests/data/input/test_image.tif
+++ b/tests/data/input/test_image.tif
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:2adf13410f2bb93ad0bbe693c5e2288c590e6935f4be812c2234faf755780b44
+size 58788
--- a/tests/test_redact_images.py
+++ b/tests/test_redact_images.py
@@ -1,41 +1,28 @@
 from pathlib import Path
 
-import tifftools
+from click.testing import CliRunner
+import pytest
 
-import imagedephi
-
-
-def verify_redaction_one_tag(ifd, tag, rule):
-    if rule['method'] == imagedephi.RedactMethod.REPLACE:
-        assert ifd['tags'][tag.value]['data'] == rule['replace_value']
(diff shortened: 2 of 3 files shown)


brianhelba (changes requested):
I'm worried that these tests contain too much complex logic. We should try to put precondition setup into fixtures. Also, we should try to make easy test case have only one logical thing it's asserting; if we need to test different things, it's fine to just make multiple test cases, each of which calls the function under test in a different way.

---

Additionally, unit tests of internal functions, especially ones with tightly coupled assertions (like asserting call counts) make it more burdensome to refactor code in the future, as the fragile tests will need to be updated too.

For example, if this is an end-to-end test, then is it absolutely essential that `redact_images` calls `redact_one_image` exactly once, or is it just fine to ensure that as a black box system, the input image practically gets redacted? If this is a unit test, are you not already convinced that your code currently calls `redact_one_image` and/or are you worried that a future regression will cause it to not be called (and that such an assertion is the only way to detect a future breakage)?

---

In this PR, I'd propose that we start with just one end-to-end test of the whole CLI. Call it with some typical input arguments, and ensure that it produces output in the expected location with the expected (i.e. redacted) properties.

Once we get this in place, we can later add more end-to-end tests for expected failure / edge cases and other new features.

Let's talk about internal unit testing strategies a bit later. Code coverage is a pathological metric (knowing that a line was executed isn't the same as knowing assertions were made about it), but it can still be a useful tool to find under-exercised parts of our codebase, which might be awkward to test end-to-end. Also, once our internal APIs get a little more stable and "pure" (in the mathematical sense of predictably mapping an input to an output), we can start writing unit tests as a way to increase confidence in the behavior of specific functions.

brianhelba (approved):

PR #20, opened 2022-12-13
Upgrade to Python 3.11

### LINKED ISSUE #16: Upgrade to Python 3.11
* This allows the use of `NotRequired` in `TypedDict`: https://docs.python.org/3/whatsnew/3.11.html#pep-655-marking-individual-typeddict-items-as-required-or-not-required
* It is 10-60% faster
* `TaskGroup` can be used instead of `asyncio.gather`: https://github.com/DigitalSlideArchive/ImageDePHI/blob/a937e55b454c0b0b01e7a90e8bb3e051f2949739/imagedephi/main.py#L64
* Consider replacing `asyncio.wait_for` with [`asyncio.timeout()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.timeout): https://github.com/DigitalSlideArchive/ImageDePHI/blob/a937e55b454c0b0b01e7a90e8bb3e051f2949739/tests/test_e2e.py#L52
* The [`Self` type](https://docs.python.org/3/whatsnew/3.11.html#pep-673-self-type) can be used to annotate factory methods.

brianhelba:
Fixes #16

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.black]
 line-length = 100
 skip-string-normalization = true
-target-version = ["py310"]
+target-version = ["py311"]
 
 [tool.isort]
 profile = "black"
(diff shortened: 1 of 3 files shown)


brianhelba:
The approach in #37 fixes this annotation in a strong way (it's just not as elegant).

Additionally, Pyoxidizer doesn't yet support Python 3.11 (though work is ongoing to add this).

PR #25, opened 2022-12-13
Add support for Git LFS

brianhelba:


brianhelba:
@naglepuff We don't want larger binary files to get committed directly to Git history, as this will permanently bloat our repo sizes (even if the files are later deleted) and slow down initial checkouts for everyone.

To verify that this works, could you rebase #19 on top of this branch, and redo the commit the adds the data file? You should observe that the file will be included in `git lfs ls-files --all` afterwards.

naglepuff (approved):

PR #29, opened 2022-12-16
Use PEP 621 for setuptools metadata

### LINKED ISSUE #21: Use project.toml for build metadata
Using `pyproject.toml` has significant implications about how a package is built internally.

First, we can choose to implement [PEP 518](https://peps.python.org/pep-0518/), by just adding a `[build-system]` section. This changes how setuptools gets internally involved by pip (and indirectly by Tox). In short, `python setup.py ...` no longer gets run internally, rather a series of standard function hooks are involved. The direct benefits are fewer intermediary build artifacts for theoretically faster builds and using a more modern (theoretically better supported) path though setuptools internals. It also facilitate [PEP 660](https://peps.python.org/pep-0660/) for the exact same benefits with editable installs.

Second, we can choose to implement [PEP 621](https://peps.python.org/pep-0621/) by moving the metadata from `setup.py` to `[project]`. Since this format is declarative (as opposed to arbitrary Python in `setup.py` that needs to get evaluated), it has much better validation (e.g. it would complain about the space in the package name) and much better documented (whereas the actual arguments to `setup(...)` have obscure backwards compatibility considerations to support distutils and older versions, and aren't comprehensively documented).

There would be additional considerations if we were publishing packages to PyPI or using an alternative build backend like Poetry, but we're not doing either presently.

brianhelba:
This provides stronger validation (since the information is static, not dynamically computed by executing `setup.py`). It is also the preferred format for `setuptools-scm`, which will help with dynamic versioning in the future.

Fixes #21.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,17 @@
 requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "ImageDePHI"
+version = "0.1.0"
+requires-python = ">=3.10"
+dependencies = [
+    "tifftools",
+]
+
+[project.scripts]
+imagedephi = "imagedephi:main"
+
 [tool.black]
 line-length = 100
 target-version = ["py310"]
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,0 @@
-from setuptools import setup
-
-setup(
-    name="ImageDePHI",
-    install_requires=["tifftools"],
-    entry_points={"console_scripts": ["imagedephi=imagedephi:main"]},
-)


pyproject.toml
```diff
@@ -2,6 +2,17 @@
 requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "ImageDePHI"
+version = "0.1.0"
```
brianhelba, line 7: `version` is a required field, so I'm arbitrarily choosing this number to start.

pyproject.toml
```diff
@@ -2,6 +2,17 @@
...
+requires-python = ">=3.10"
```
brianhelba, line 8: This was implicitly the Python version required, but it ought to be explicitly checked at install time.

naglepuff (approved):

PR #35, opened 2022-12-16
Switch to Click

brianhelba:


manthey:
I don't get it, @brianhelba.  You proposed this task as a learning opportunity, but then did it yourself.  It adds no functionality and pulls in an additional dependency.  Why?

PR #41, opened 2023-01-10
Web test

marySalvi:
Adds basics tests for gui and fixes a couple of bugs found with the tests: 
1. Handle non-directory selection
2. Handle non-existent path


imagedephi/gui.py
```diff
@@ -12,9 +12,12 @@
 def select_directory(request: Request, path: str = "/"):
     base_path = pathlib.Path(path)
 
-    if not base_path.is_dir():
+    if not base_path.exists():
```
brianhelba, line 15: [`Path.is_dir`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_dir) states:
> `False` is also returned if the path doesn’t exist or is a broken symlink

Why is a separate check for both `.exists()` and `.is_dir()` necessary?

tests/test_gui.py
```diff
@@ -14,3 +16,41 @@ def test_gui_home(client: TestClient):
 
     assert response.status_code == 200
     assert "Select Directory" in response.text
+
+
+def test_gui_navigate_success(
+    client: TestClient,
+    tmp_path: Path,
+):
+    response = client.get("/?path=%s" % str(tmp_path))
```
brianhelba, line 25: See https://www.python-httpx.org/quickstart/#passing-parameters-in-urls . We can make the test client encode the query parameters for us, which is more readable and ensures that URL encoding happens properly.

tests/test_gui.py
```diff
@@ -14,3 +16,41 @@ def test_gui_home(client: TestClient):
...
+    assert response.json() == {"message": "You chose this directory: %s" % tmp_path}
+
+
+def test_gui_directory_selection_failure(
+    client: TestClient,
+    tmp_path: Path,
+):
+    response = client.post("/directory_selection/", data={"directory": str(tmp_path / "fake")})
+
+    assert response.status_code == 404
```
brianhelba, line 56: Do you care to make an assertion about the response body here? I don't care either way, I just noticed that you did so for the `test_gui_navigate_not_found` case.

brianhelba (commented):
Great improvement to test coverage, I think this is exactly the right scope and structure of tests for these endpoints.

Some minor questions / comments inline.

imagedephi/gui.py
```diff
@@ -12,9 +12,12 @@
 def select_directory(request: Request, path: str = "/"):
     base_path = pathlib.Path(path)
```
brianhelba, line 13:
```suggestion
def select_directory(request: Request, path: pathlib.Path = pathlib.Path("/")):
```

We should be able to make FastAPI do the type coercion for us as it parses the query string.

imagedephi/gui.py
```diff
@@ -28,4 +31,8 @@ def select_directory(request: Request, path: str = "/"):
 
 @app.post("/directory_selection/")
 def selection(directory: str = Form()):  # noqa: B008
```
brianhelba, line 33:
```suggestion
def selection(directory: pathlib.Path = Form()):  # noqa: B008
```

We should be able to make FastAPI do the type coercion for us as it parses the form. You could then remove the explicit coercion on the next line.

tests/test_gui.py
marySalvi: updated in cbdc6e68684f30990e1b2f5e8764d59ac3fbbebf

tests/test_gui.py
marySalvi: updated in cbdc6e68684f30990e1b2f5e8764d59ac3fbbebf

imagedephi/gui.py
marySalvi: fixed in cbdc6e68684f30990e1b2f5e8764d59ac3fbbebf

imagedephi/gui.py
marySalvi: fixed in cbdc6e68684f30990e1b2f5e8764d59ac3fbbebf

imagedephi/gui.py
marySalvi: It's not. I was mistaken.

brianhelba (approved):
LGTM. Please fix the type checking errors, then go ahead and merge it

PR #42, opened 2023-01-13
Implement command for redacting from rulesets

naglepuff:
This PR explores transforming rule definitions (likely to be specified in a yaml file) into a series of rules.

The new objects and classes are simplified such that they only really work when trying to redact TIFF metadata based on TIFF tags.

Some basic rule sets have been added, mostly for the purposes of testing. `base_rules.yaml` contains rules that cover all tiff tags for some images that I use for testing. `example_user_rules.yml` contains a single rule for `ImageDescription`, which can be used to override the corresponding rule from `base_rules.yaml`.

To demonstrate how these new classes could be useful, there are two new commands:

`imagedephi redaction-plan <image> <override_rules>`: This will compose the base and override rulesets specified, and build a redaction plan based on the image provided. If TIFF metadata tags are found in the image with no corresponding rule in the composed ruleset, that is reported.

`imagedephi run <input dir> <output dir> <override_rules>`: This has been updated to use the `TiffMetadataRedactionPlan` class to handle redaction of data.


imagedephi/main.py
```diff
@@ -2,15 +2,18 @@
 
 import asyncio
 from pathlib import Path
+import pkgutil
+from typing import Optional, TextIO
```
brianhelba, line 6: [`typing.Optional` is unnecessary](https://docs.python.org/3/library/typing.html#typing.Optional) (and unofficially deprecated) in Python 3.10. Just write `| None`.

imagedephi/main.py
```diff
@@ -24,12 +27,31 @@ def imagedephi() -> None:
     "input-dir", type=click.Path(exists=True, file_okay=False, readable=True, path_type=Path)
 )
 @click.argument(
-    "output-dir", type=click.Path(exists=True, file_okay=False, writable=True, path_type=Path)
+    "output-dir", type=click.Path(exists=True, file_okay=False, readable=True, path_type=Path)
```
brianhelba, line 30: If it needs to be both readable and writable, I'd suggest explicitly adding both kwargs.

imagedephi/main.py
```diff
@@ -24,12 +27,31 @@ def imagedephi() -> None:
...
 )
-def run(input_dir: Path, output_dir: Path) -> None:
-    """Run in CLI-only mode."""
-    redact_images(input_dir, output_dir)
-    click.echo("Done!")
+@click.argument("override-rules", type=click.File("r"), required=False)
```
brianhelba, line 32: This should be a `click.option`: https://click.palletsprojects.com/en/8.1.x/options/ , which will be more user-friendly than an optional positional argument.

imagedephi/main.py
```diff
@@ -24,12 +27,31 @@ def imagedephi() -> None:
...
+def run(input_dir: Path, output_dir: Path, override_rules: Optional[TextIO]):
+    """Redact images in a folder according to given rule sets."""
+    base_rules_bytes = pkgutil.get_data("imagedephi", "base_rules.yaml")
```
brianhelba, line 35: The better-supported way to do this is to use [`importlib.resources.files`](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files), which supports implicit namespace packages (which means you don't have to have an `__init__.py` in a resource folder for it to be loadable).

```suggestion
    base_rules_path = importlib.resources.files("imagedephi") / "base_rules.yaml"
```

In this case, `base_rules_path` is a [`Traversable`](https://docs.python.org/3.10/library/importlib.html#importlib.abc.Traversable), which supports basically a subset of the `pathlib.Path` API. So, you can:
```python
with base_rules_path.open() as base_rules_stream:
    yaml.safe_load(base_rules_stream)
```

imagedephi/main.py
```diff
@@ -24,12 +27,31 @@ def imagedephi() -> None:
...
+    if base_rules_bytes is None:
+        raise click.ClickException("There was an error with finding the base rules.")
```
brianhelba, line 37:
```suggestion
```

If this happens, something is catastrophically wrong with our packaging, so it's fine to let errors naturally get thrown by whatever is actually trying to do the reading.

imagedephi/main.py
```diff
@@ -24,12 +27,31 @@ def imagedephi() -> None:
...
+    override_rule_set = build_ruleset(yaml.safe_load(override_rules)) if override_rules else None
+    redact_images(input_dir, output_dir, base_rule_set, override_rule_set)
+
+
+@imagedephi.command
+@click.argument("image", type=click.Path())
+@click.argument("override-rules", type=click.File("r"), required=False)
+def redaction_plan(image: click.Path, override_rules: Optional[TextIO]) -> None:
+    """Print the redaction plan for a given image and rules."""
+    click.echo(override_rules)
```
brianhelba, line 48: Rather than implement all this rule-loading logic in every sub-command, we can add it to the root command, then pass the parsed data as Click "context" to every subcommand. See https://click.palletsprojects.com/en/8.1.x/complex/

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@dataclass
+class TiffTagEntry:
+    tag: tifftools.TiffTag
+    entry: TagEntry
+
+
+class RuleType(Enum):
+    METADATA = 1
```
brianhelba, line 20:
```suggestion
    METADATA = "metadata"
```

Rather than arbitrary numbers, how about string values for all these `Enum`?

naglepuff: Didn't notice you changed the recommendation to use `auto()`. Would that be preferred to the strings? I updated to strings in e5d83e6.

imagedephi/main.py
naglepuff: Makes sense to me. This was removed as part of transitioning to `importlib.resources`. (c234aa9)

imagedephi/main.py
naglepuff: Thanks for the pointer. Implemented this change in c234aa9

imagedephi/main.py
naglepuff: 7178ba5

imagedephi/main.py
naglepuff: a66dc07 ensures that the output dir is writable. I think we want both, right? We want to know what files already exist, in case we need to overwrite, and we need to be able to write the redacted images to the directory.

imagedephi/main.py
naglepuff: 79fdcc4
Noted that `typing.Optional` is deprecated.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+@dataclass
+class RuleSet:
+    name: str
+    description: str
+    rules: dict[RuleFormat, list[Rule]]
+
+
+def _make_tiff_metadata_rule(rule: dict) -> TiffMetadataRule:
+    """Transform a rule from schema into an object."""
+    tag = tifftools.constants.Tag[rule["tag"]]
```
brianhelba, line 79: `tifftools.constants.Tag` also supports lookup by tag name. For example: `tifftools.constants.Tag['ImageDescription']`. If the tag name isn't known, we can catch a `KeyError`, and the rule would be invalid.

I think it'd be a more friendly rule syntax if we could typically specify rules by tag name.

For example:
```yaml
    - type: metadata
      tag_name: ImageWidth
      method: keep
```

As a fallback, we could also support:
```yaml
    - type: metadata
      tag_id: 256
      method: keep
```

But I think it'd be fine if we start with `tag_name` only.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class RuleSource(Enum):
+    BASE = "base"
+    OVERRIDE = "override"
+
+
+@dataclass
+class Rule:
+    title: str
```
brianhelba, line 41: I'd suggest making `title` optional, and actually naming it `description`. In most cases, I believe we have enough information from the rest of the rule to auto-generate the printed rule description (which an explcit `description` value would override). This would typically make rules more maintainable and less likely to contain contradictory content, since I suspect a lot of people will be copy-pasting them.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import TYPE_CHECKING
+
+import tifftools
+
+if TYPE_CHECKING:
+    from tifftools.tifftools import IFD, TagEntry
+
+
+@dataclass
+class TiffTagEntry:
```
brianhelba, line 14: Is this used anywhere?

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+    TIFF = "tiff"
+
+
+class RuleSource(Enum):
+    BASE = "base"
+    OVERRIDE = "override"
+
+
+@dataclass
+class Rule:
```
brianhelba, line 40: Could we just add a `rule_source` within every `Rule` instance? I see that you want to know it later for printing / debugging purposes, so it seems easier to just be able to look it up directly from the rule.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_title(self) -> str:
+        """Generate a title for the rule."""
+        return self.title
+
+
+@dataclass
+class TiffMetadataRule(Rule):
+    tag: tifftools.TiffTag
+    replace_value: str | bytes | list[int | float]
```
brianhelba, line 56: I think we should allow `None` here, for anything other than `RedactMethod.REPLACE`. You could also further subclass this `RedactTiffMetadataRule`, but maybe that's too complicated.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+
+def _make_tiff_metadata_rule(rule: dict) -> TiffMetadataRule:
+    """Transform a rule from schema into an object."""
+    tag = tifftools.constants.Tag[rule["tag"]]
+    return TiffMetadataRule(
+        title=rule["title"],
+        redact_method=RedactMethod[rule["method"].upper()],
+        rule_type=RuleType.METADATA,
+        tag=tag,
+        replace_value=rule.get("new_value", ""),
```
brianhelba, line 85: You could have some more validation here, like:
```suggestion
        replace_value=rule["new_value"] if redact_method == RedactMethod.REPLACE else None
```

Ultimately, we could use a JSON schema or just Pydantic to validate the rule file.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+
+_rule_function_mapping = {RuleFormat.TIFF: {RuleType.METADATA: _make_tiff_metadata_rule}}
+
+
+def make_rule(rule_format: RuleFormat, rule_type: RuleType, rule: dict) -> Rule:
+    rule_function = _rule_function_mapping[rule_format][rule_type]
+    return rule_function(rule)
```
brianhelba, line 94: IMO, this is a bit too complicated. I think the `build_ruleset` factory function should be moved to this file, then I think you should complement it with another factory function which just looks like:
```
def _build_rule(rule_format: RuleFormat, rule_dict: dict) -> Rule:
```

That could replace all of `_make_tiff_metadata_rule`, `_rule_function_mapping`, and `make_rule`. It's fine for `_build_rule` to use explicit `if ... elif` for now, just like `TiffMetadataRule.apply`. I'm worried that the current approach is just too hard to understand, for a premature benefit. It some point a dynamic lookup will probably make sense, but a lot of things could change until then.

Btw, I don't care about the name `make_, `build_`, etc., but it should probably be consistent between `build_ruleset` and `build_rule`.

brianhelba: Instead of:
```python
def _build_rule(rule_format: RuleFormat, rule_dict: dict) -> Rule:
```
you could also have
```python
class TiffMetadataRule:

    @classmethod
    def build(rule_dict: dict) -> TiffMetadataRule:
```
for the factory. That should probably help to make the type bounds a bit stronger too.

imagedephi/rules.py
brianhelba: Feel free to disregard this if you think it makes constructing rules a lot more complicated.

imagedephi/redact.py
```diff
@@ -1,74 +1,166 @@
...
+        ]
+        if override_rules is None:
+            self.override_rules = []
+        else:
+            self.override_rules = [
+                rule
+                for rule in override_rules.rules[RuleFormat.TIFF]
+                if isinstance(rule, TiffMetadataRule)
+            ]
+        self._build_redaction_steps(self.image_data["ifds"])
```
brianhelba, line 83: It seems like you could inline `_build_redaction_steps` without making things much more complicated here. If this class ends up being that difficult to construct, I think we should try to make it simple.

imagedephi/rules.py
```diff
@@ -0,0 +1,94 @@
...
+    IMAGE = "image"
+
+
+class RedactMethod(Enum):
+    REPLACE = "replace"
+    DELETE = "delete"
+    KEEP = "keep"
+
+
+class RuleFormat(Enum):
```
brianhelba, line 30: This name isn't intuitive to me. Maybe call it `FileFormat`?

imagedephi/redact.py
```diff
@@ -1,74 +1,166 @@
...
+    def __init__(
+        self, tiff_info: TiffInfo, base_rules: RuleSet, override_rules: RuleSet | None
+    ) -> None:
+        self.redaction_steps = {}
+        self.no_match_tags = []
+        self.image_data = tiff_info
+        self.base_rules = [
+            rule for rule in base_rules.rules[RuleFormat.TIFF] if isinstance(rule, TiffMetadataRule)
+        ]
+        if override_rules is None:
```
brianhelba, line 75: Is there value in this class "knowing" that `override_rules` is non-existent? If not, just make `override_rules` be required as a list, and make it the responsibility of the caller to pass an empty list. That will simplify a lot of this logic.

imagedephi/redact.py
```diff
@@ -1,74 +1,166 @@
...
+            self.override_rules = []
+        else:
+            self.override_rules = [
+                rule
+                for rule in override_rules.rules[RuleFormat.TIFF]
+                if isinstance(rule, TiffMetadataRule)
+            ]
```
brianhelba, line 82: Since there are no other rule types now, I think this filtering is premature, and it adds unnecessary code complexity. If we did have other rule types, lots of other things would probably fail anyway. I'd suggest just assigning it directly.

naglepuff: I do like the idea of leaving out this sort of filtering for now. Perhaps `RuleSet.rules` could have a type of:
`dict[FileType, dict[RuleType, list[Rule]]`

The problem here is that `TiffMetadataRedactionPlan` expects its `base_rules` to be a list of `TiffMetadataRules`, so directly assigning makes mypy mad.

imagedephi/rules.py
naglepuff: I agree, this makes a lot of sense. Rules knowing their own source could be helpful. Additionally it gets rid of an ad-hoc tuple data structure. Refactored in 2fe7513

imagedephi/rules.py
naglepuff: I implemented this in 50a1284. It led to a typing issue that I wasn't quite sure of the best solution to. We can't set the value of a tiff entry to `None`, but if a rule is using the `replace` method, and `replace_value` is `None`, that's what we'd be doing. 

We don't have validation on rules yet, so theoretically this could happen. My approach in the aforementioned commit is to just throw a runtime error, but maybe this should be walked back, and we just put in an empty string by default?

We could start doing some rule validation, and quit out early if we detect invalid override rules. Open to suggestions.

imagedephi/rules.py
naglepuff: d5b6d2d refactors rule construction. I think this gets closer to what you're imagining. I particularly think that the `TiffMetadataRule.build` class method is good. Using `if` statements instead of the indirection I think is also more clear.

imagedephi/redact.py
naglepuff: Inlined both this function and `_redact_image` in 95f43bf. The use of the generator to iterate over IFDs made these functions perhaps too small to justify being functions themselves.

imagedephi/rules.py
naglepuff: Agreed. Changed in b3b1bd7

imagedephi/redact.py
naglepuff: See 19899e1 for changes here.

I've changed how the `TiffMetadataRedactionPlan` constructor works to simplify it, so the relevant logic is now elsewhere. The caller now passes a list of `TiffMetadataRules`, so your suggestion has been implemented.

imagedephi/redact.py
naglepuff: See 19899e1 for changes here.

This filtering has been moved to the `RuleSet.get_tiff_metadata_rules` method. I *think* the filtering is necessary because the `RuleSet.rules` object contains lists of generic rules, so `mypy` doesn't like treating the list as a list of `TiffMetadataRules`.

I have changed that structure slightly so that all `TiffMetadataRule`s can live in their own list, instead of sharing with the future `TiffImageRule` list. 

Open to suggestions here.

imagedephi/rules.py
naglepuff: Seeing the rules written out with `tag_name` definitely makes me think this is a better starting point than using the number. It's much easier to read. 4b06806 switches it over.

imagedephi/rules.py
naglepuff: Agreed. See 676431c

naglepuff:
@brianhelba I've gone over the changes we discussed at the end of last week. I think they all make sense, so I'm going to merge the changes.

PR #43, opened 2023-01-17
Output directory selector

marySalvi:
Allows a user to traverse two directory trees independently and select an input and output directory for images:

Screenshot from 2023-01-17 16-23-50
Screenshot from 2023-01-17 16-24-06
Screenshot from 2023-01-17 16-24-28


brianhelba (approved):
:shipit:

PR #49, opened 2023-01-26
Gui display files

marySalvi:
Displays supported images(svs and tif) in directory list.

image

### DIFF
--- a/imagedephi/gui.py
+++ b/imagedephi/gui.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import asyncio
+from dataclasses import dataclass
 from pathlib import Path
+from typing import List
 
 from fastapi import BackgroundTasks, FastAPI, Form, HTTPException, Request
 from fastapi.responses import HTMLResponse
@@ -13,6 +15,28 @@ app = FastAPI()
 templates = Jinja2Templates(directory=Path(__file__).parent / "templates")
 
 shutdown_event = asyncio.Event()
+
+
+@dataclass
+class DirectoryData:
+    directory_path: Path
+    images: List[Path]
+
+
+def _create_directory_list(path: Path) -> List[DirectoryData]:
+    directory_list: List[DirectoryData] = []
+    for directory in path.iterdir():
+        if directory.is_dir():
+            children: List[Path] = []
+            try:
+                for child in directory.iterdir():
+                    if child.suffix == ".tif" or child.suffix == ".svs":
+                        children.append(child)
+            except PermissionError:
+                pass
+            directory_list.append({"directory_path": directory, "images": children})
+
+    return directory_list
 
 
 @app.on_event("startup")
@@ -38,36 +62,12 @@ def select_directory(
     output_bread_crumbs = list(reversed(output_directory.parents))
     output_bread_crumbs.append(output_directory)
 
-    # input_directories = [
-    #     path_element for path_element in input_directory.iterdir() if path_element.is_dir()
-    # ]
-    # output_directories = [
-    #     path_element for path_element in output_directory.iterdir() if path_element.is_dir()
-    # ]
-
-    def createDirectoryList(path):
-        directory_list = []
-        for directory in path.iterdir():
-            if directory.is_dir():
-                children = []
-                if directory.is_dir():
-
-                    try:
-                        for child in directory.iterdir():
-                            if child.suffix == ".tif" or child.suffix == ".svs":
-                                children.append(child)
-                    except PermissionError:
-                        pass
-                directory_list.append({"directory_path": directory, "images": children})
-
-        return directory_list
-
     return templates.TemplateResponse(
         "DirectorySelector.html.j2",
         {
             "request": request,
-            "input_directories": createDirectoryList(input_directory),
-            "output_directories": createDirectoryList(output_directory),
+            "input_directories": _create_directory_list(input_directory),
+            "output_directories": _create_directory_list(output_directory),
             "input_bread_crumbs": input_bread_crumbs,
             "output_bread_crumbs": output_bread_crumbs,
             "current_input": input_directory,
--- a/imagedephi/gui.py
+++ b/imagedephi/gui.py
@@ -18,21 +18,22 @@ shutdown_event = asyncio.Event()
 
 @dataclass
 class DirectoryData:
-    directory_path: Path
+    directory_list: list[Path]
     images: list[Path]
 
 
-def _create_directory_list(path: Path) -> list[DirectoryData]:
-    directory_list: list[DirectoryData] = []
+def _create_directory_list(path: Path) -> DirectoryData:
+    path_info: DirectoryData = DirectoryData(directory_list=[], images=[])
+
     for directory in path.iterdir():
         if directory.is_dir():
-            try:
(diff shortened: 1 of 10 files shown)


imagedephi/redact.py
```diff
@@ -114,32 +114,43 @@ def get_base_rules():
...
+
 def redact_images(
     image_dir: Path,
     output_dir: Path,
     override_rules: RuleSet | None = None,
     overwrite: bool = False,
 ) -> None:
     base_rules = get_base_rules()
-    for child in image_dir.iterdir():
+    for image_file in iter_image_files(image_dir):
```
marySalvi, line 135: @naglepuff note the refactor done here.

manthey:
Perhaps foolishly, I tried this on a directory where I have LOTS of files (a spot where I had sub directories with ~8000 files in one directory and lots in others).  I had to search for my directory name.  I would have expected that I would drill all the way into a directory before showing files -- and then if there were a lot it wouldn't necessarily load them all.  Once I navigate all the way into a directory, it seems odd that I have to navigate back up to select the directory -- I'd expect that the CHOOSE DIRECTORIES would choose the directories I am in rather than selecting a directory and then choosing it.

manthey:
Can we please sort the directories (at least when it starts in my root directory, the order seems arbitrary)?

marySalvi:
> Perhaps foolishly, I tried this on a directory where I have LOTS of files (a spot where I had sub directories with ~8000 files in one directory and lots in others). I had to search for my directory name. I would have expected that I would drill all the way into a directory before showing files -- and then if there were a lot it wouldn't necessarily load them all. Once I navigate all the way into a directory, it seems odd that I have to navigate back up to select the directory -- I'd expect that the CHOOSE DIRECTORIES would choose the directories I am in rather than selecting a directory and then choosing it.

Perhaps change to click for select, double click to navigate into the directory?

manthey:
> Perhaps change to click for select, double click to navigate into the directory?

Unless you can choose multiple directories, I think the current behavior is surprising.  When I've navigated into a directory, there isn't any way to select it -- you have to navigate back out.  Perhaps the "CHOOSE DIRECTORIES" button should be "use these directories", and it would use the directories you are in for input and output.  Selecting a directory would still enter it and show its contents; there wouldn't be any radio button, since the directory listed would be the selected directory.  For instance, in this screenshot:
image
my initial assumption was that it would use /mnt/data3/TCGA and /tmp, and if I wanted to use /mnt/data3/TCGA/square and /tmp/r, I should click each of those directories.  This would change the file list to just be the current directory (those in /mnt/data/TCGA rather than those in each of the directories in /mnt/data/TCGA), which would be more obvious to me what is about to be operated on, since it is all of the images you see rather than a subset.

imagedephi/redact.py
```diff
@@ -114,32 +114,43 @@ def get_base_rules():
...
 
 
+def iter_image_files(directory: Path) -> Generator[Path, None, None]:
+    """
+    Given a directory return an iterable of available images.
+
+    May raise a PermissionError if the directory is not readable.
+    """
+    for child in directory.iterdir():
+        if child.suffix == ".tif" or child.suffix == ".svs":
```
manthey, line 124: We should check if it is in `FILE_EXTENSION_MAP`.

jeffbaumes:
> > Perhaps change to click for select, double click to navigate into the directory?
> 
> Unless you can choose multiple directories, I think the current behavior is surprising. When I've navigated into a directory, there isn't any way to select it -- you have to navigate back out. Perhaps the "CHOOSE DIRECTORIES" button should be "use these directories", and it would use the directories you are in for input and output. Selecting a directory would still enter it and show its contents; there wouldn't be any radio button, since the directory listed would be the selected directory. For instance, in this screenshot: image my initial assumption was that it would use /mnt/data3/TCGA and /tmp, and if I wanted to use /mnt/data3/TCGA/square and /tmp/r, I should click each of those directories. This would change the file list to just be the current directory (those in /mnt/data/TCGA rather than those in each of the directories in /mnt/data/TCGA), which would be more obvious to me what is about to be operated on, since it is all of the images you see rather than a subset.

I think @manthey is on the right track, where a directory with no subdirectory selected means "select the current folder". Google Drive's "move" dialog has similar behavior, where the "move" button changes name to "move here" to distinguish between these states. Since we have two folders to select that button name change doesn't make sense. Note that Google Drive also allows you to deselect a child folder by clicking on it again to essentially reselect the parent.

Feb-16-2023 09-32-12

Similarly Google Drive's folder upload opens Mac's folder selection dialog, which seems to always have an active "Upload" button whether the user has chosen a folder or not. In the latter case it's selecting the folder the user has navigated to. It does not seem the Mac dialog easily allows folder deselection (you need to command-click to do it).

We don't need to follow these exactly, but in general I feel it's good to follow a UX pattern from somewhere else when we can, kind of like deciding to use a library, we can leverage knowledge from others instead of reinventing. There may be other "folder pickers" out there to consider.

imagedephi/redact.py
marySalvi: Are there additional extensions I should include in the map or just add as we go?

manthey: For now, we have a very limited tool that only works on tiff files and variations.  There are lots of extensions that we've seen for tiff files (ptif for Philips' generated tiff files, for instance).  From a UI perspective, the existing file extension map is file.  Eventually, we'll want to be able to let the user filter by their own method (including trying all files), and not care about file extension.

The more correct approach is to use something like the "magic" library to detect file format, or, more certain, use large_image and its tile sources -- but this requires reading some of each file and is therefore slow, especially in cases where we have thousands of files in a directory.

marySalvi:
> Perhaps foolishly, I tried this on a directory where I have LOTS of files (a spot where I had sub directories with ~8000 files in one directory and lots in others). I had to search for my directory name. I would have expected that I would drill all the way into a directory before showing files -- and then if there were a lot it wouldn't necessarily load them all. Once I navigate all the way into a directory, it seems odd that I have to navigate back up to select the directory -- I'd expect that the CHOOSE DIRECTORIES would choose the directories I am in rather than selecting a directory and then choosing it.


fixed in 34f43a442f1b267ec0acef726e0fab93f7387bc1 and 30949e8046d2b11e2713536dbf34874e13b1849e

marySalvi:
> Can we please sort the directories (at least when it starts in my root directory, the order seems arbitrary)?

fixed in 366b107145e10569ad4261726b5e240487e3331f

imagedephi/gui.py
```diff
@@ -27,40 +44,25 @@ def select_directory(
     input_directory: Path = Path("/"),  # noqa: B008
     output_directory: Path = Path("/"),  # noqa: B008
 ):
+    Path("")
```
brianhelba, line 47: @brianhelba This shouldn't be committed, remove it.

marySalvi: removed in d6e4e2812808229709879d48721e30756d7fc1bd

imagedephi/redact.py
marySalvi: filed as issue #61 to fix after merge if #55  is not merged first

brianhelba (approved):

PR #50, opened 2023-01-26
Handle case where previous output exists

### LINKED ISSUE #48: Handle existing file in output
Currently if you run redaction on an image and it already exists in the output directory it fails silently. First pass could just be informing the user that it failed but would probably be ideal to give the user options of renaming, overwriting or canceling.

naglepuff:
Fixes #48 

A new flag has been added to the `run` command to allow overwriting existing output in the output directory.

Ex: `imagedephi run images/ output/ -o`

**Previous Behavior**
The `run` command would fail silently when attempting to write to an existing file.

**New Behavior**
Running `imagedephi run` without the new flag will not write over existing output images. For each image it cannot save due to existing files, it will print a message for the user informing them of what is happening.

Running `imagedephi run` with the new flag will overwrite existing redacted images, and also report to the user that it is doing so.

### DIFF
--- a/imagedephi/redact.py
+++ b/imagedephi/redact.py
@@ -94,13 +94,6 @@ def _get_output_path(file_path: Path, output_dir: Path) -> Path:
     return output_dir / f"REDACTED_{file_path.name}"
 
 
-def get_base_rules():
-    base_rules_path = importlib.resources.files("imagedephi") / "base_rules.yaml"
-    with base_rules_path.open() as base_rules_stream:
-        base_rule_set = build_ruleset(yaml.safe_load(base_rules_stream), RuleSource.BASE)
-        return base_rule_set
-
-
 def _save_redacted_tiff(tiff_info: TiffInfo, output_path: Path, input_path: Path, overwrite: bool):
     if overwrite or not output_path.exists():
         if output_path.exists():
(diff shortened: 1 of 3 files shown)


imagedephi/redact.py
```diff
@@ -131,8 +131,24 @@ def _get_output_path(file_path: Path, output_dir: Path) -> Path:
...
+def _save_redacted_tiff(tiff_info: TiffInfo, output_path: Path, input_path: Path, overwrite: bool):
+    if overwrite or not output_path.exists():
+        if output_path.exists():
+            click.echo(f"Found existing redaction for {input_path.name}. Overwriting...")
+        tifftools.write_tiff(tiff_info, output_path, allowExisting=True)
+    else:
+        click.echo(
+            f"Could not redact {input_path.name}, existing redacted file in output directory. "
+            "Use the --overwrite-existing-output flag to overwrite previously redacted files."
+        )
```
brianhelba, line 143:
```suggestion
    if output_path.exists():
        if overwrite:
            click.echo(f"Found existing redaction for {input_path.name}. Overwriting...")
        else:
           click.echo(
               f"Could not redact {input_path.name}, existing redacted file in output directory. "
               "Use the --overwrite-existing-output flag to overwrite previously redacted files."
           )
           return
    tifftools.write_tiff(tiff_info, output_path, allowExisting=True)
```

I think this logic structure is a bit easier to follow, but it's up to you.

brianhelba (approved):
See comment for a possible improvement. Otherwise, LGTM.

imagedephi/redact.py
```diff
@@ -131,8 +131,24 @@ def _get_output_path(file_path: Path, output_dir: Path) -> Path:
     return output_dir / f"REDACTED_{file_path.name}"
 
 
+def _save_redacted_tiff(tiff_info: TiffInfo, output_path: Path, input_path: Path, overwrite: bool):
+    if overwrite or not output_path.exists():
+        if output_path.exists():
+            click.echo(f"Found existing redaction for {input_path.name}. Overwriting...")
+        tifftools.write_tiff(tiff_info, output_path, allowExisting=True)
+    else:
+        click.echo(
```
manthey, line 140: Shouldn't we be adding `err=True` to `click.echo` calls that are errors?

naglepuff: https://github.com/DigitalSlideArchive/ImageDePHI/issues/52

I created this issue to track this concern. Will look into this as a follow up.

PR #55, opened 2023-02-10
Support SVS Images

### LINKED ISSUE #53: Support redaction for Aperio/svs format
Aperio/svs files make special use of the ImageDescription tiff tag.

So far, our approach to redaction has been to remove the ImageDescription from tiffs by default. Doing this for svs files will invalidate files as svs. At a minimum we need to add logic to instead modify the ImageDescription value to keep all data before the first pipe ("|") character, and keep the AppMag and mpp key-value pairs.

naglepuff:
Closes #53 

This PR creates additional classes and adds logic to add support for redacting [SVS/Aperio format whole slide images](https://openslide.org/formats/aperio/). It does this by following a similar path to plain tiff redaction, but treating the `ImageDescription` tag in svs IFD differently.

TODO

- [x] Create test images and write tests for SVS files


naglepuff:
@manthey @brianhelba @marySalvi 
This is in a mostly working state, so feel free to test on any .svs images you have around. I will be out until 2/20, but in the meantime please take a look at these changes and offer any feedback you have. When I get back we can discuss everything in more detail.

(If your images happen to have metadata properties that aren't covered in `base_rules.yaml`, you can create a rule override file and add any rules needed to actually get this to run on your images.)

manthey:
I ran this on some images and those images do open with ImageScope (so Aperio thinks they are still Aperio files).

I'll run it on the fraction of TCGA slides I have and see where it wants more rules.

imagedephi/redact.py
```diff
@@ -90,6 +161,96 @@ def execute_plan(self) -> None:
...
+        self,
+        tiff_info: TiffInfo,
+        base_rules_tiff: list[MetadataTiffRule],
+        base_rules_svs: list[MetadataSvsRule],
+        override_rules_tiff: list[MetadataTiffRule],
+        override_rules_svs: list[MetadataSvsRule],
+    ) -> None:
+        super().__init__(tiff_info, base_rules_tiff, override_rules_tiff)
+
+        image_description_tag = tifftools.constants.Tag["ImageDescription"]
```
manthey, line 187: We have some mislabeled or malformed tiff files that do NOT have ImageDescription tags.  These fail:
```
  File "/home/manthey/misc/ImageDePHI/imagedephi/redact.py", line 74, in build
    return SvsMetadataRedactionPlan(
  File "/home/manthey/misc/ImageDePHI/imagedephi/redact.py", line 188, in __init__
    del self.redaction_steps[image_description_tag.value]
KeyError: 270
```
We should more gracefully refuse to process them (or handle them as plain tiff files).

manthey:
We really need a method saying to keep this if it matches a pattern (like a number) otherwise redact it.

imagedephi/redact.py
```diff
@@ -90,6 +161,96 @@ def execute_plan(self) -> None:
...
+                )
+                for tag in self.no_match_tags:
+                    click.echo(f"{tag.value} - {tag.name}")
+            if self.no_match_description_keys:
+                click.echo(
+                    "The following keys were found in Aperio ImageDescription strings "
+                    "and could not be redacted given the current set of rules."
+                )
+                for key in self.no_match_description_keys:
+                    click.echo(key)
```
manthey, line 225: It'd be nice to make this output more grepable -- when I run on a huge list of files it'd be nice to have "Missing tag, Aperio: {key}" to filter on. (and "Missing tag, tiff: {key}" in the tiff output).

Also, I don't know why we are using `click.echo` instead of `print`.  We aren't using *any* of the coloring or other features of click, so this just makes the code harder to read.

imagedephi/redact.py
```diff
@@ -67,8 +135,11 @@ def __init__(
...
 
+    def is_comprehensive(self) -> bool:
+        return len(self.no_match_tags) == 0
+
     def report_missing_rules(self) -> None:
-        if self.no_match_tags:
+        if not self.is_comprehensive():
             click.echo("The following tags could not be redacted given the current set of rules.")
             for tag in self.no_match_tags:
                 click.echo(f"{tag.value} - {tag.name}")
```
manthey, line 145: `{tag.name}` is printing as a numerical value for me, so I see:
```
The following tags could not be redacted given the current set of rules.
305 - 305
```

manthey:
I ran this across a fairly large collection of TIFF and SVS files and the following tags were unhandled:
[unhandled.txt](https://github.com/DigitalSlideArchive/ImageDePHI/files/10761211/unhandled.txt)

imagedephi/redact.py
naglepuff: dfb3610 makes use of this parameter.

imagedephi/redact.py
```diff
@@ -11,20 +12,87 @@
...
     Represents a plan of action for redacting metadata from TIFF images.
 
     The plan can be used for reporting to end users what steps will be made to redact their TIFF
     images, and also executing the plan.
     """
 
+    file_format = FileFormat.TIFF
     tiff_info: TiffInfo
     redaction_steps: dict[int, MetadataTiffRule]
     no_match_tags: list[tifftools.TiffTag]
```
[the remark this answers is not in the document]
naglepuff: Done as part of dfb3610

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: GPSLatitudeRef
+      method: delete
+      type: metadata
+    - tag_name: OldSubfileType
+      method: delete
+      type: metadata
+    - tag_name: Threshholding
+      method: delete
+      type: metadata
+    - tag_name: FillOrder
```
manthey, line 109: Don't delete FillOrder.  It is needed for decoding.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: MinSampleValue
+      method: delete
+      type: metadata
+    - tag_name: MaxSampleValue
+      method: delete
+      type: metadata
+    - tag_name: PageName
+      method: delete
+      type: metadata
+    - tag_name: Xposition
```
manthey, line 130: Don't delete XPosition or Yposition (again, we'd ideally check that these are numerical values).

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+      method: delete
+      type: metadata
+    - tag_name: Yposition
+      method: delete
+      type: metadata
+    - tag_name: T4Options
```
manthey, line 136: Don't delete T4Options or T6Options; they are needed for decoding faxes.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+      method: delete
+      type: metadata
+    - tag_name: T6Options
+      method: delete
+      type: metadata
+    - tag_name: PageNumber
```
manthey, line 142: Don't delete PageNumber.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+      method: delete
+      type: metadata
+    - tag_name: GPSDateStamp
+      method: delete
+      type: metadata
+    - tag_name: GPSLatitude
+      method: delete
+      type: metadata
+    - tag_name: '306'
```
manthey, line 151: Is this distinct from DateTime?

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: DateTime
+      method: delete
+      type: metadata
+    - tag_name: Artist
+      method: delete
+      type: metadata
+    - tag_name: HostComputer
+      method: delete
+      type: metadata
+    - tag_name: ColorMap
```
manthey, line 163: Don't delete ColorMap.  It is needed for palettized images.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+      method: delete
+      type: metadata
+    - tag_name: HalftoneHints
```
manthey, line 166: We probably want to keep Halftonehints, and the various Fax tags.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: ConsecutiveBadFaxLines
+      method: delete
+      type: metadata
+    - tag_name: Matteing
+      method: delete
+      type: metadata
+    - tag_name: Datatype
+      method: delete
+      type: metadata
+    - tag_name: TileDepth
```
manthey, line 184: We probably want to keep TileDepth -- I'm not sure which images use it.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: InkSet
+      method: delete
+      type: metadata
+    - tag_name: InkNames
+      method: delete
+      type: metadata
+    - tag_name: Copyright
+      method: delete
+      type: metadata
+    - tag_name: ExposureTime
```
manthey, line 196: We want to keep ExposureTime and FNumber.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: GeoDoubleParamsTag
+      method: delete
+      type: metadata
+    - tag_name: GeoAsciiParamsTag
+      method: delete
+      type: metadata
+    - tag_name: ExposureProgram
+      method: delete
+      type: metadata
+    - tag_name: ISOSpeedRatings
```
manthey, line 247: Let's keep ISOSpeedRatings.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: ExifVersion
+      method: delete
+      type: metadata
+    - tag_name: DateTimeOriginal
+      method: delete
+      type: metadata
+    - tag_name: CreateDate
+      method: delete
+      type: metadata
+    - tag_name: ShutterSpeedValue
```
manthey, line 262: Let's keep the values from ShutterSpeedValue through FocalLength.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: GPSLongitudeRef
+      method: delete
+      type: metadata
+    - tag_name: FlashpixVersion
+      method: delete
+      type: metadata
+    - tag_name: '40961'
+      method: delete
+      type: metadata
+    - tag_name: ColorSpace
```
manthey, line 319: Keep.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+      method: delete
+      type: metadata
+    - tag_name: PixelXDimension
```
manthey, line 322: Keep these dimensions, the focal plane values, and the sensing method.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: FocalPlaneResolutionUnit
+      method: delete
+      type: metadata
+    - tag_name: SensingMethod
+      method: delete
+      type: metadata
+    - tag_name: FileSource
+      method: delete
+      type: metadata
+    - tag_name: SceneType
```
manthey, line 343: Keep SceneType through SubjectDistanceRange.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: '50839'
+      method: delete
+      type: metadata
+    - tag_name: '51123'
+      method: delete
+      type: metadata
+    - tag_name: '51168'
+      method: delete
+      type: metadata
+    - tag_name: JPEGProc
```
manthey, line 400: Keep JPEGProc through YCbCrPositioning.

imagedephi/base_rules.yaml
```diff
@@ -82,3 +82,615 @@ rules:
...
+    - tag_name: '60011'
+      method: delete
+      type: metadata
+    - tag_name: '60020'
+      method: delete
+      type: metadata
+    - tag_name: '65200'
+      method: delete
+      type: metadata
+    - tag_name: NDPI_ByteCountHighBytes
```
manthey, line 466: We shouldn't include ANY NDPI tags in our base rule set.  Sadly, NDPI files are *nearly* tiff files, but they aren't tiff files and handling them as tiff files will produce broken hdmi files.

imagedephi/base_rules.yaml
naglepuff: It doesn't appear to be. The tiff standard doesn't mention anything else it could be, and there's only one tag in `tifftools.constants` with that number. I wrote a python script to generate rules from `unhandled.txt` and it looks like there was an extra line in that file, resulting in a redundant rule.

```
306 - DateTime
306 - 306
```

manthey: And I generated the unhandled.txt file with some horrible bashisms, so it is probably my fault it is in there...

manthey (approved):

brianhelba (approved):

PR #62, opened 2023-03-14
Add Pydantic model for tiff rules

marySalvi:
Adds pydantic model for tiff and svs rules.


imagedephi/models/rules_models.py
```diff
@@ -0,0 +1,41 @@
...
+class Keep(BaseModel):
+    method: Literal["keep"]
+
+
+class Delete(BaseModel):
+    method: Literal["delete"]
+
+
+class Rule(BaseModel):
+    description: str | None = None
```
marySalvi, line 22: @brianhelba we discussed default `None` vs truly optional. This thread does offer some interesting solutions but I'm not sure if any would work for us.
https://github.com/pydantic/pydantic/issues/1223#issuecomment-589986747

imagedephi/models/rules_models.py
```diff
@@ -0,0 +1,41 @@
...
+    method: Literal["replace"]
+    new_value: str
+
+
+class Keep(BaseModel):
+    method: Literal["keep"]
+
+
+class Delete(BaseModel):
+    method: Literal["delete"]
```
marySalvi, line 18: @brianhelba these would likely be removed. They were part of the attempt to use the discriminator, but unless we intend to make method a nested dict I don't think this structure will work for us.

PR #64, opened 2023-03-17
Reorganize rules and redaction code

naglepuff:
This breaks out `redact.py` and `rules.py` into two modules: `redact` and `rules`.


imagedephi/redact/tiff.py
```diff
@@ -0,0 +1,128 @@
...
+    def build(
+        cls, image_path: Path, base_rules: RuleSet, override_rules: RuleSet | None = None
+    ) -> TiffBasedMetadataRedactionPlan:
+        file_extension = FILE_EXTENSION_MAP[image_path.suffix]
+        for redaction_plan_class in cls.get_all_subclasses():
+            if file_extension == redaction_plan_class.file_format:
+                tiff_info = tifftools.read_tiff(str(image_path))
+                return redaction_plan_class(tiff_info, base_rules, override_rules)
+        else:
+            raise Exception(f"File format for {image_path} not supported.")
```
naglepuff, line 41: @brianhelba This is slightly different. Previously the `build` class method would do hard checks against members of the `FileFormat` enum. Now it traverses its subclasses. Because of the current class structure, it also needs to check all of its grandchildren, hence the addition of the `get_all_subclasses` method.

naglepuff: This also meant that I had to unify the signatures of the `__init__` functions on `TiffMetadataRedactionPlan` and `SvsMetadataRedactionPlan`.

imagedephi/redact.py
```diff
@@ -1,348 +0,0 @@
...
-            )
-        if FILE_EXTENSION_MAP[file_extension] == FileFormat.SVS:
-            tiff_info = tifftools.read_tiff(str(image_path))
-            return SvsMetadataRedactionPlan(
-                tiff_info,
-                base_rules.get_metadata_tiff_rules(),
-                base_rules.get_metadata_svs_rules(),
-                override_rules.get_metadata_tiff_rules() if override_rules else [],
-                override_rules.get_metadata_svs_rules() if override_rules else [],
-            )
```
naglepuff, line 86: Notice how the two redaction plan classes have different signatures. This was changed as part of the reorg.

imagedephi/redact/__init__.py
```diff
@@ -0,0 +1 @@
+from .redact import iter_image_files, redact_images, show_redaction_plan  # noqa
```
brianhelba, line 1: You add [an `__all__` in this file](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) so that the `# noqa` isn't necessary.

Alternatively, you could make this file blank, and refactor imports to reach into the sub-module files.

brianhelba: Or at least use `#noqa: ERRCODE`

brianhelba (approved):

PR #76, opened 2023-03-21
Launch the GUI when run directly from Windows Explorer

brianhelba:


brianhelba:
I'm going to add a unit test to try and figure out the failure on Windows.

brianhelba:
Something in GitHub is wrong with this PR. Opening #90 to replace this.

PR #79, opened 2023-03-29
Prototype for type-checking rules

naglepuff:
This PR introduces one possible way to incorporate type-checking into the redaction process. It adds a new rule type called `datatype`. Rules of this new type should always have a property called `expected_type`. In the example rules added in this PR, I've added one such rule with an `expected_type` of `number` (open to discussion of what they should be - should the allowed values here match the DataTypes defined in `tifftools`?)


naglepuff:
Closing in favor of #126

PR #82, opened 2023-03-31
Allow Windows users to get help via "/?"

### LINKED ISSUE #75: Windows help via /?
Windows command line program frequently have `/?` as an option to show help.  One the main command, this will show help because it doesn't match anything, but on subcommands it matches a root directory whose name is one character long, which is definitely not what a windows user would expect.

brianhelba:
Fixes #75.


imagedephi/main.py
```diff
@@ -24,10 +25,18 @@ class ImagedephiContext:
     override_rule_set: RuleSet | None = None
 
 
+CONTEXT_SETTINGS = {"help_option_names": ["--help"]}
+if sys.platform == "win32":
+    # Allow Windows users to get help via "/?".
+    # To avoid ambiguity with actual paths, only support this on Windows.
```
manthey, line 31: The other option would be to check if it is an actual path and use it; if not, give help.

brianhelba: I want to avoid these sort of decision trees in application logic.

On Windows, this will indeed mean it's impossible to specify an input path of `/?`, but I think it's exceedingly unlike that such a path would ever exist.

PR #85, opened 2023-04-05
Reorganize utility functions

brianhelba:


tests/test_e2e.py
```diff
@@ -45,7 +45,7 @@ def test_e2e_run(runner: CliRunner, data_dir: Path, tmp_path: Path) -> None:
 
 
 @pytest.mark.timeout(5)
-def test_e2e_plan(runner: CliRunner, data_dir: Path, tmp_path: Path) -> None:
+def test_e2e_plan(runner: CliRunner, data_dir: Path) -> None:
```
brianhelba, line 48: Removed unnecessary use of `tmp_path`.

tests/test_utils_network.py
```diff
@@ -0,0 +1,26 @@
...
+    server = await asyncio.start_server(server_callback, "127.0.0.1", unused_tcp_port)
+    async with server:
+        yield server
+        server.sockets[0]
+
+
+@pytest.mark.timeout(1)
+@pytest.mark.asyncio
+async def test_utils_network_wait_for_port(server: asyncio.Server) -> None:
+    server_port = server.sockets[0].getsockname()[1]
```
brianhelba, line 24: Get `server_port` directly from `server`, instead of relying on the fact that `unused_tcp_port` is shared between the test case and the fixture (which is true, but makes the test dependent on the implementation of the fixture).

brianhelba:
This is primarily a just a reorganization, but #86 depends on it, so I'm merging this now.

PR #86, opened 2023-04-05
Use a random port for serving the GUI

### LINKED ISSUE #73: If GUI port is in use and not explicitly specified, pick a different port automatically

brianhelba:
Fixes #73.

### DIFF
--- a/imagedephi/utils/network.py
+++ b/imagedephi/utils/network.py
@@ -1,4 +1,5 @@
 import asyncio
+import socket
 
 
 async def wait_for_port(port: int, host: str = "127.0.0.1") -> None:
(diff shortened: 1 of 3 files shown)


brianhelba:
@marySalvi PTAL.

imagedephi/utils/network.py
```diff
@@ -12,3 +13,14 @@ async def wait_for_port(port: int, host: str = "127.0.0.1") -> None:
             writer.close()
             await writer.wait_closed()
             return
+
+
+def unused_tcp_port() -> int:
+    with socket.socket() as sock:
+        sock = socket.socket()
```
marySalvi, line 20: Just for my own understanding, wouldn't sock already be `socket.socket()` from line 19? I am unsure why line 20 is necessary.

marySalvi (approved):
LGTM Tested on windows and ubuntu

imagedephi/utils/network.py
brianhelba: You're right. I think I was testing something and accidently committed this line. I'll fix it.

marySalvi (approved):

PR #88, opened 2023-04-07
Gui handle existing output

marySalvi:
Provides a UI for feature parity with the backend when handling an existing output file.
[Screencast from 04-07-2023 12:35:11 PM.webm](https://user-images.githubusercontent.com/40494088/230645457-6de23e29-9691-4a95-a1b3-d9c6fef4b65f.webm)


marySalvi:
@naglepuff 
I wonder if this:
https://github.com/DigitalSlideArchive/ImageDePHI/blob/38b394513aba5960e281eda299ace888dd41f63b/imagedephi/redact/redact.py#L32-L35
should be moved here:
https://github.com/DigitalSlideArchive/ImageDePHI/blob/38b394513aba5960e281eda299ace888dd41f63b/imagedephi/main.py#L63-L64


Just a thought, I'm not sure where it's best handled and whether its good practice to ignore and exception or not.

brianhelba:
@naglepuff I agree with your general concern. Since this is a feature design question, I'm going to re-open #48 and discuss it more there.

brianhelba (changes requested):
I'd like to reduce the amount of duplicated code in this PR, but I want to make sure that we're working on the right problem. Let's discuss the design in #48, then I can leave comments on the code.

marySalvi:
Closing in favor of new approach outlined in #48. Will delete branch when new PR is incorporated.

PR #90, opened 2023-04-18
Launch the GUI when run directly from Windows Explorer

brianhelba:
On Windows, this will cause the ImageDePHI GUI to start when a user runs `imagedephi.exe` directly from Windows Explorer. In all other cases (non-Windows and Windows CLI usage), the existing behavior (running with no arguments prints help text) is unchanged.

This is implemented via a new `FallthroughGroup` subclass of `click.Group`. This implementation abstracts the logic of falling through out of the main CLI endpoints (which are more concerned with handling arguments). The decoupled `FallthroughGroup` is also easier to test.

A new `launched_from_windows_explorer` utility uses Windows platform APIs to determine whether the current ImageDePHI process is running via a Windows command line shell or was launched directly by Windows Explorer. See code comments for implementation details.

### DIFF
--- a/imagedephi/utils/cli.py
+++ b/imagedephi/utils/cli.py
@@ -1,7 +1,9 @@
 import asyncio
 from collections.abc import Callable, Coroutine
 from functools import wraps
-from typing import ParamSpec, TypeVar
+from typing import Any, ParamSpec, TypeVar
+
+import click
 
 T = TypeVar("T")
 P = ParamSpec("P")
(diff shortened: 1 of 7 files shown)


imagedephi/utils/cli.py
```diff
@@ -15,3 +17,32 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
...
+    """A Group which may run a subcommand when no subcommand is specified."""
+
+    def __init__(self, subcommand_name: str, should_fallthrough: Callable[[], bool], **attrs: Any):
+        # Subcommands are not added until after this is instantiated,
+        # so only store the future subcommand name
+        self.subcommand_name = subcommand_name
+        self.should_fallthrough = should_fallthrough
+
+        attrs["invoke_without_command"] = True
+        attrs["no_args_is_help"] = False
```
naglepuff, line 32: Just making sure I understand - this subclass of `click.Group` behaves the same as the stock `click.Group`, except if its being run from the windows explorer, it interprets no arguments as running "fallthrough" command instead of just printing the help text?

tests/test_utils_cli.py
```diff
@@ -13,3 +15,54 @@ def test_utils_cli_run_coroutine(mocker: MockerFixture) -> None:
...
+        subcommand_name="sub", should_fallthrough=lambda: True, callback=cmd
+    )
+    cmd_group.add_command(click.Command(name="sub", callback=sub))
+
+    result = cli_runner.invoke(cmd_group, [])
+
+    assert result.exit_code == 0
+    cmd.assert_called_once()
+    sub.assert_called_once()
+    assert "Usage" not in result.output
```
naglepuff, line 68: Could these tests be parametrized? Maybe with new parameters like  `should_fallthrough_func` and `expected_call_count`?

naglepuff:
Could you also add a description to this PR?

marySalvi (approved):
LGTM. Tested in virtual machine and it works as expected. Thanks for linking the resources. I really enjoyed learning how this works. Very interesting.

tests/test_utils_cli.py
brianhelba: Done. The call count is actually not essential to the behavior. It's sufficient to expect that the function was called at all or not.

imagedephi/utils/cli.py
brianhelba: Yes, but the class is more generic, which makes it easier to test and reason about. If there are no arguments and the value of the `should_fallthrough` is true, then it runs the specified subcommand instead of printing the help text.

imagedephi/utils/os.py
```diff
@@ -0,0 +1,35 @@
...
+        # https://learn.microsoft.com/en-us/windows/console/getconsoleprocesslist
+        # indicate that the "process_list" array may not be null.
+        # Also "process_list" must have a size larger than 0, but its full content isn't needed.
+        process_list_size = 1
+        # Array elements should be DWORD, which is a uint
+        process_list = (ctypes.c_uint * process_list_size)()
+        process_count: int = ctypes.windll.kernel32.GetConsoleProcessList(
+            process_list, process_list_size
+        )
+        if process_count == 0:
```
brianhelba, line 27: Note, I've made this into an explicit exception, rather than just returning `False`. We should always expect this API to succeed on Windows. If it doesn't, we want to hear about it.

PR #93, opened 2023-04-19
Move to dict-style rules

naglepuff:


imagedephi/rules/ruleset.py
```diff
@@ -1,49 +1,88 @@
...
+                key_name
+            ] = SvsFormatRules.build_svs_image_description_rule(key_name, method_info, rule_source)
+
+
+@dataclass
+class RuleSet:
+    name: str
+    description: str
+    tiff: TiffFormatRules
+    svs: SvsFormatRules
```
naglepuff, line 67: This, combined with the new classes defined above, unify the structure of the RuleSet class and the shape of the rule set files.

imagedephi/rules/ruleset.py
```diff
@@ -1,49 +1,88 @@
...
 
-    raise Exception("Unsupported rule.")
+    def get_format_rules(self, file_format: FileFormat) -> FileFormatRules:
+        if file_format == FileFormat.TIFF:
+            return self.tiff
+        if file_format == FileFormat.SVS:
+            return self.svs
+        raise Exception(f"File format {file_format} not supported")
```
naglepuff, line 74: Since this returns the generic `FileFormatRules` class, type checking in consuming code may be affected. Is there a better way to do this?

imagedephi/rules/ruleset.py
```diff
@@ -1,49 +1,88 @@
...
+    rule_set_args = [ruleset_spec["name"], ruleset_spec["description"]]
+    for file_format in FileFormat:
+        format_rules = ruleset_spec.get(file_format.value, None)
+        if not format_rules:
+            raise Exception(f"No rules provided for format {file_format}")
+        if file_format == FileFormat.TIFF:
+            rule_set_args.append(TiffFormatRules(format_rules, rule_source))
+        elif file_format == FileFormat.SVS:
+            rule_set_args.append(SvsFormatRules(format_rules, rule_source))
+    return RuleSet(*rule_set_args)
```
naglepuff, line 88: I should rework this to avoid having to pass an args list to the `RuleSet` constructor. Mypy was giving me a tough time because the tiff/svs rules may not be defined by the time we hit the return statement.

imagedephi/redact/tiff.py
```diff
@@ -90,16 +89,19 @@ def __init__(
...
+        merged_rules = base_rules | override_rules if override_rules else base_rules
         for tag, _ in self._iter_tiff_tag_entries(ifds):
             # First iterate through overrides, then base
-            for rule in chain(override_rules, base_rules):
-                if rule.is_match(tag):
-                    self.redaction_steps[tag.value] = rule
-                    break
+            tag_rule = merged_rules.get(tag.value, None)
+            if tag_rule and tag_rule.is_match(tag):
+                self.redaction_steps[tag.value] = tag_rule
```
naglepuff, line 103: Merging dictionaries instead of chaining lists for the updated shape of rulesets.

naglepuff:
Closing in favor of #95, which has the same changes but cleaner git history.

PR #94, opened 2023-04-21
Create output file

### LINKED ISSUE #48: Handle existing file in output

### LINKED ISSUE #72: Have run create output directory if it doesn't exist
Maybe there should be an option regarding this?

marySalvi:
- Makes `output_dir` flag optional
- Removes overwrite flag
- Creates a new directory at the supplied `output_dir` named `Redacted_[time_stamp]` and places redacted images inside. 
- Defaults to current directory if no `output_dir` is given.

Closes #48 Closes #72

### DIFF
--- a/imagedephi/main.py
+++ b/imagedephi/main.py
@@ -64,19 +64,20 @@ def imagedephi(ctx: click.Context, override_rules: TextIO | None) -> None:
     default=Path('.'),
     type=click.Path(exists=True, file_okay=False, readable=True, writable=True, path_type=Path),
 )
-@click.option(
-    "-o",
-    "--overwrite-existing-output",
-    is_flag=True,
-    default=False,
-    help="Overwrite previous output for input images.",
-)
+# I don't think we need this anymore
+# @click.option(
+#     "-o",
+#     "--overwrite-existing-output",
+#     is_flag=True,
+#     default=False,
+#     help="Overwrite previous output for input images",
+# )
 @click.pass_obj
 def run(
-    obj: ImagedephiContext, input_path: Path, output_dir: Path, overwrite_existing_output: bool
+    obj: ImagedephiContext, input_path: Path, output_dir: Path
 ):
     """Perform the redaction of images."""
-    redact_images(input_path, output_dir, obj.override_rule_set, overwrite_existing_output)
+    redact_images(input_path, output_dir, obj.override_rule_set)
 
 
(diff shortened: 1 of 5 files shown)


marySalvi:
Opened as a draft for implementation discussion. Will need to adjust tests if this is the agreed implementation. 
I believe this eliminates the need of a new template for the gui.

@manthey @naglepuff @brianhelba

imagedephi/main.py
```diff
@@ -45,21 +45,24 @@ def imagedephi(ctx: click.Context, override_rules: TextIO | None) -> None:
 @click.argument("input-path", type=click.Path(exists=True, readable=True, path_type=Path))
 @click.argument(
     "output-dir",
+    required=False,
+    default=Path('.'),
```
brianhelba, line 49: https://docs.python.org/3/library/pathlib.html#pathlib.Path.cwd

imagedephi/main.py
```diff
@@ -45,21 +45,24 @@ def imagedephi(ctx: click.Context, override_rules: TextIO | None) -> None:
 @click.argument("input-path", type=click.Path(exists=True, readable=True, path_type=Path))
 @click.argument(
     "output-dir",
+    required=False,
```
brianhelba, line 48: This is actually an opportunity to make the CLI take [variadic input arguments](https://click.palletsprojects.com/en/8.1.x/arguments/#variadic-arguments). This would allow it to work with shell expansion or multiple named files.

If we want that, we should make the output directory into an option, instead of an argument.

brianhelba: @marySalvi Brought up the point that this change would require some additional internal logic. Let's do it in a follow-up PR.

tests/test_e2e.py
```diff
@@ -95,7 +98,8 @@ def client_select_directory(port: int) -> httpx.Response:
 
     assert cli_result.exit_code == 0
     webbrowser_open_mock.assert_called_once()
-    output_file = tmp_path / "REDACTED_test_image.tif"
+    time_stamp = datetime.datetime.now().isoformat(timespec="seconds")
```
brianhelba, line 101: https://github.com/spulec/freezegun

PR #95, opened 2023-04-26
Move to dict-style rules

naglepuff:
Co-authored-by: Brian Helba <brianhelba@users.noreply.github.com>

This PR uses new pydantic models for rules, removing the `imagedephi.rules` module in the process. It is mostly a swap between the new models and the old rule classes. Some of the functionality found in the old rule classes has been moved, as described in conversations below.

This simplifies loading/building rulesets from yaml files, and makes it easier to find errors in rules yaml.


imagedephi/redact/tiff.py
```diff
@@ -82,26 +82,46 @@ def _iter_tiff_tag_entries(
...
         self.redaction_steps = {}
         self.no_match_tags = []
         ifds = self.tiff_info["ifds"]
-        override_rules = override_rule_set.get_metadata_tiff_rules() if override_rule_set else []
-        base_rules = base_rule_set.get_metadata_tiff_rules()
+        override_rules = (
+            override_rule_set.get_format_rules(self.file_format).metadata
+            if override_rule_set
+            else None
+        )
```
naglepuff, line 97: Since `TiffMetadataRedactionPlan.__init__` can be called by subclasses, I wanted a way to get the right format rules from a `Ruleset`, which is why `Ruleset.get_format_rules` exists. It does feel kind of hacky, so if there's a better way to accomplish this, please let me know. @brianhelba

imagedephi/redact/tiff.py
```diff
@@ -82,26 +82,46 @@ def _iter_tiff_tag_entries(
...
+            rule_tag = get_tiff_tag(rule.key_name)
+            return rule_tag.value == tag.value
+        return False
+
+    def apply(self, rule: ConcreteMetadataRule, ifd: IFD) -> None:
+        tag = get_tiff_tag(rule.key_name)
+        if rule.action == "delete":
+            del ifd["tags"][tag.value]
+        elif rule.action == "replace":
+            ifd["tags"][tag.value]["data"] = rule.new_value
```
naglepuff, line 123: Note that this functionality has been moved from the old `Rule` classes to the `RedactionPlan` classes. If this change feels like the concept of a `RedactionPlan` is becoming bloated, I'd like to discuss potential designs that separate these concerns better.

imagedephi/redact/tiff.py
```diff
@@ -115,8 +135,8 @@ def report_missing_rules(self) -> None:
 
     def report_plan(self) -> None:
         print("Tiff Metadata Redaction Plan\n")
-        for rule in self.redaction_steps.values():
-            print(rule.get_description())
+        for tag_value, rule in self.redaction_steps.items():
+            print(f"Tiff Tag {tag_value} - {rule.key_name}: {rule.action}")
         self.report_missing_rules()
```
naglepuff, line 140: Because the file format is now tied to the `RedactionPlan` alone, and no longer the individual rules, I've moved the logic of printing out rules to the `RedactionPlan` classes, but if another home makes sense (like the `TiffRules` and `SvsRules` classes), let me know.

stubs/tifftools/constants.pyi
```diff
@@ -28,6 +28,7 @@ class TiffConstantSet(Generic[_TiffConstantT]):
     ) -> _TiffConstantT | None: ...
 
 class TiffTag(TiffConstant):
+    altnames: set[str] | None
```
naglepuff, line 31: See https://github.com/DigitalSlideArchive/tifftools/blob/0c39e95119c302f553d818260a28cbd83a7da98e/tifftools/constants.py#L627 as an example of a tag with multiple names.

imagedephi/models/rules.py
```diff
@@ -0,0 +1,88 @@
...
+            return self.tiff
+        if file_format == FileFormat.SVS:
+            return self.svs
+        raise Exception("File format not supported")
+
+
+base_rules_path = importlib.resources.files("imagedephi") / "base_rules.yaml"
+
+with base_rules_path.open() as stream:
+    rules_model = Ruleset.parse_obj(yaml.safe_load(stream))
```
brianhelba, line 88: What does this do? Why execute this at import-time?

imagedephi/models/rules.py
```diff
@@ -0,0 +1,88 @@
+from enum import Enum
```
brianhelba, line 1: You can feel free to put this into whatever file you'd like. Maybe it should go back into `rules/`?

imagedephi/models/rules.py
naglepuff: This doesn't need to be here. It might have been for testing at some point? It has been removed.

imagedephi/models/rules.py
naglepuff: The `rules` module was removed. Is there an issue with defining the enum here?

naglepuff:
@brianhelba sorry to add some additional bloat to this PR, but I decided to get rid of the intermediate class `TiffBasedMetdataRedactionPlan`. The builder function is now in its own file. I believe that change plus the changes we discussed earlier today make things a lot cleaner.

brianhelba (approved):
LGTM :shipit:

PR #96, opened 2023-04-26
Update readme

marySalvi:
Moves the development instructions to `DEVELOPMENT.md` updates the readme to include installation and usage instructions.


marySalvi:
I added everyone to review in case there is anything glaringly wrong or missing

DEVELOPMENT.md
```diff
@@ -0,0 +1,19 @@
+# ImageDePHI
+
+This project has been funded in whole or in part with Federal funds from the National Cancer Institute, National Institutes of Health, Department of Health and Human Services, under Contract No. 75N91022C00033
```
brianhelba, line 3: This should stay in the main `README`.

DEVELOPMENT.md
```diff
@@ -0,0 +1,19 @@
...
+
+## To Run as a CLI
+
+`pip install -r requirements.txt`
+
+`python imagedephi.py <directory of input images> <directory to store redacted images>`
```
brianhelba, line 9: I don't think we want developers doing any of this now.

For local development, we probably want them to:
* create a virtualenv (we should assume they know how to dot his already, but we could maybe link to docs)
* run `pip install -e .`
* execute with `imagedephi`

For running tests, I think we can just tell people:
* `pip install tox`
* `tox`

If we explicitly document how to run tests, I think we should actually remove the `requirements.txt` file, since it only contains `tox`.

README.md
```diff
@@ -1,19 +1,53 @@
...
 
-`python imagedephi.py <directory of input images> <directory to store redacted images>`
+[ImageDePHI Artifacts](https://github.com/DigitalSlideArchive/ImageDePHI/actions/workflows/ci.yml?query=branch%3Amain)
 
+Select the most recent run:
+![ci run](imagedephi/assets/ci_run.png)
 
-## Development
+Then select the appropriate artifact for your OS:
+![artifacts](imagedephi/assets/artifacts_list.png)
```
brianhelba, line 15: I think we can have a change very soon which just publishes binaries directly with release. Asking people to go into CI runs is too burdensome just to download an application.

README.md
```diff
@@ -1,19 +1,53 @@
...
+Commands:
+  gui   Run a web-based GUI.
+  plan  Print the redaction plan for a given image and rules.
+  run   Redact images in a folder according to given rule sets.
+```
+
+To redact a set of images supply the directory of the images to be processed and [optionally] an output directory to deposit the redacted images:
+```
+imagedephi run [path/to/images] [optional/output/path]
+```
```
brianhelba, line 46: I don't think we should include this output. It's going to be changing frequently, it may be slightly different on different platforms (e.g. #75), and it will be burdensome for us to keep these docs up to date.

README.md
```diff
@@ -1,19 +1,53 @@
...
+
+## GUI
+Alternatively you can use the gui to navigate to the desired directories:
```
brianhelba, line 49:
```suggestion
Alternatively you can use the GUI to navigate to the desired directories:
```

README.md
```diff
@@ -1,19 +1,53 @@
...
+```
+imagedephi gui
 ```
+### **Note**: Windows users can double click the imagdephi executable icon to launch the gui outside of the terminal workflow
```
brianhelba, line 53:
```suggestion
### **Note**: To launch the ImageDePHI GUI, Windows users can directly run `imagdephi.exe` from Windows Explorer.
```

DEVELOPMENT.md
marySalvi: updated in a89e9c9d71a8b2ec6a0ddaa81f819ffd05d1f6e7

README.md
marySalvi: Once we have a release I will update this to link to that instead

README.md
marySalvi: removed in a89e9c9d71a8b2ec6a0ddaa81f819ffd05d1f6e7

README.md
marySalvi: fixed in a89e9c9d71a8b2ec6a0ddaa81f819ffd05d1f6e7

README.md
marySalvi: fixed in a89e9c9d71a8b2ec6a0ddaa81f819ffd05d1f6e7

README.md
```diff
@@ -1,19 +1,34 @@
...
+ImageDePHI is an application that can be used to redact PHI from whole slide images.
+
 
 This project has been funded in whole or in part with Federal funds from the National Cancer Institute, National Institutes of Health, Department of Health and Human Services, under Contract No. 75N91022C00033
 
-## To Run as a CLI
 
-`pip install -r requirements.txt`
+# Installation
+ImageDePHI is under active development and as such the best way to test the latest features is to download the most current artifact.
```
brianhelba, line 9: https://github.com/DigitalSlideArchive/ImageDePHI/releases/latest will take you to the latest release

README.md
```diff
@@ -1,19 +1,34 @@
 # ImageDePHI
+ImageDePHI is an application that can be used to redact PHI from whole slide images.
```
brianhelba, line 2:
```suggestion
ImageDePHI is an application to redact PHI from whole slide images.
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,44 @@
+## Development
```
brianhelba, line 1: Put in `docs/development.md`

DEVELOPMENT.md
```diff
@@ -0,0 +1,44 @@
...
+* create a virtualenv: [venv docs](https://docs.python.org/3/library/venv.html)
```
brianhelba, line 2:
```suggestion
* create and activate a virtualenv: [venv docs](https://docs.python.org/3/library/venv.html)
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,44 @@
...
+* run :
+    ```
```
brianhelba, line 4:
```suggestion
    ```bash
```

brianhelba:
@marySalvi This content looks good. I've just pushed an additional commit to clean up the style a bit more, to make it publication quality.

brianhelba (approved):

PR #102, opened 2023-05-05
Redact associated images

naglepuff:
This PR is a first pass at image redaction for tiff and aperio files.

The default rules have been updated as follows:
1. Tiff associated images - keep all. For generic tiff images where we can't determine a more specific format, just keep all associated images. 
2. SVS label image - replace with black image. Aperio files need this associated image to be read by Aperio software.
3. SVS macro - delete. Prevent the case where the label bleeds onto the macro
4. Thumbnail + unidentifiable - keep

I'm open to suggestions for better default behavior.

For replacing images, this PR brings in PIL as a dependency. PIL is used here to create black images, which can be saved and read by PIL as a TIFF to get replacement IFDs.


imagedephi/base_rules.yaml
```diff
@@ -3,6 +3,9 @@ description: These are the default rules for each potential piece of redactable
   (metadata, image) to be used as a fallback if there are no user defined rules for
   a piece of redactable material.
 tiff:
+  associated_images:
+    default:
+      action: delete
```
naglepuff, line 8: @manthey @brianhelba I imagine it will be a bit trickier to come up with rules for associated images, as different formats will each implement this in a different way. Does it make sense to have a default action to take on associated images? 

For example, in aperio files, the image description for an IFD can be used to determine if the image is a macro or label image (but not thumbnail). If we're processing aperio files that, for some reason, have associated images that we can't consistently identify, how should those be handled? Currently I'm taking a naive approach with the "default" rule, but I'm open to other ideas about this.

imagedephi/models/rules.py
```diff
@@ -11,27 +11,38 @@ class FileFormat(Enum):
...
-class KeepMetadataRule(_MetadataRule):
-    action: Literal["keep"]
+class ReplaceMetadataRule(ReplaceRule):
+    new_value: str
 
 
-class DeleteMetadataRule(_MetadataRule):
-    action: Literal["delete"]
+class ReplaceImageRule(_Rule):
+    replace_with: Literal["black_square"]
```
naglepuff, line 37: Found an issue. This should inherit from `ReplaceRule`.

I imagine that `keep` and `delete` rules will work the same for all images and metadata, as there's no room for interpretation. However, `ReplaceRule`s will be different depending on what they are applied to (replace metadata with string, replace an image with a different image, etc.)

imagedephi/redact/svs.py
```diff
@@ -151,3 +151,35 @@ def execute_plan(self) -> None:
...
+    file_format = FileFormat.SVS
+
+    def get_associated_image_key_for_ifd(self, ifd: IFD):
+        """Attempt to match an associated image with a rule. Fallback to default rule."""
+        image_description = str(ifd["tags"][270]["data"])
+        # we could do additional checks, like look for a macro based on dimensions
+        for key in self.final_rules:
+            if key in image_description:
+                return key
+        return "default"
```
naglepuff, line 166: @manthey is there a better way to do determine if an IFD in an svs file is a macro or label? I remember you saying that macro images are usually much wider than they are tall, due to the dimensions of the slides themselves.

I also remember you saying that if svs files don't have a macro, they won't work properly with Aperio software, so I want to make sure we properly identify macro images as often as possible.

imagedephi/redact/tiff.py
```diff
@@ -1,83 +1,130 @@
...
     def build(
-        cls, image_path: Path, base_rules: Ruleset, override_rules: Ruleset | None = None
-    ) -> TiffBasedMetadataRedactionPlan:
-        file_extension = FILE_EXTENSION_MAP[image_path.suffix]
+        cls,
+        tiff_info: TiffInfo,
+        file_format: FileFormat,
+        base_rules: Ruleset,
+        override_rules: Ruleset | None = None,
+    ) -> Self:
```
naglepuff, line 69: Previously, each redaction plan opened the file independently. Now the `tiff_info` gets passed directly so multiple plans can operate on the same image data.

imagedephi/base_rules.yaml
manthey: large_image has a bunch of heuristics to determine the appropriate label of an associated image.  Some readers (openslide) report this (mostly); some, like tiff, require that we scrape the information from the various ways different writers do things; others, like bioformats, may not expose enough information to actually determine the appropriate label.

In general, we will probably have some heuristics that will work some of the time.  When we don't know, the best we can probably do is have "image-1", "image-2", etc. without knowing what they are.  So, yes, generally, this means we will want a rule for otherwise unspecified associated images that would be delete or keep.

imagedephi/redact/svs.py
naglepuff: Thanks for the correction. As part of this PR I've added new base rules, including replace-with-blank-image for label images in svs files.

imagedephi/redact/tiff.py
naglepuff: This is no longer relevant. Instead of having multiple redaction plans, one for metadata and one for images, they are rolled into one.

imagedephi/redact/tiff.py
```diff
@@ -27,9 +35,37 @@ class TiffMetadataRedactionPlan(RedactionPlan):
...
     tiff_info: TiffInfo
-    redaction_steps: dict[int, ConcreteMetadataRule]
+    metadata_redaction_steps: dict[int, ConcreteMetadataRule]
+    image_redaction_steps: dict[int, ConcreteImageRule]
     no_match_tags: list[tifftools.TiffTag]
 
+    @staticmethod
+    def is_tiled(ifd: IFD):
+        """Determine if an IFD represents a tiled image."""
+        tile_width_tag = tifftools.constants.get_or_create_tag("TileWidth", tifftools.constants.Tag)
```
manthey, line 45: Why do you call `get_or_create_tag`?  Can't you just do `tile_width_tag = tifftools.Tag.TileWidth`?

imagedephi/redact/tiff.py
```diff
@@ -100,17 +148,78 @@ def report_missing_rules(self) -> None:
...
+            length = int(ifd["tags"][257]["data"][0])
+            output_path = temp_dir / str(uuid.uuid4())
+            image = Image.new("RGB", (width, length))
+            image.save(output_path, "TIFF", compression="jpeg")
+            return output_path
+        raise Exception("Redaction option not currently supported")
+
+    def update_new_ifd(self, old_ifd: IFD, new_ifd: IFD) -> None:
+        for tag_value, entry in old_ifd["tags"].items():
+            if tag_value not in new_ifd["tags"].keys():
```
manthey, line 173: We probably don't want to copy all tags from the old ifd that aren't in the new ifds.  You could have tags that affect the image decoding that PIL doesn't write to the new ifd (for instance, orientation could exist on the old image but not on the new; or an icc color profile might exist on the old image).  We probably only want to copy old tags that are Datatype.ASCII.

naglepuff: I'll modify the loop so it only copies over tags that are ASCII.

imagedephi/redact/tiff.py
```diff
@@ -100,17 +148,78 @@ def report_missing_rules(self) -> None:
...
+                new_entry: TagEntry = {
+                    "datatype": entry["datatype"],
+                    "count": entry["count"],
+                    "data": entry["data"],
+                }
+                new_ifd["tags"][tag_value] = new_entry
```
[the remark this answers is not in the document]
naglepuff: I'll modify th eloop to account for this as well.

stubs/tifftools/tifftools.pyi
```diff
@@ -27,7 +27,7 @@ class IFD(TypedDict):
 class TagEntry(TypedDict):
     datatype: int
     count: int
-    datapos: int
+    datapos: NotRequired[int]
```
brianhelba, line 30: It seems like this is actually always returned: https://github.com/DigitalSlideArchive/tifftools/blob/7b94211693800f997205b1d1225ef63ae894e29c/tifftools/tifftools.py#L183

When is it missing?

imagedephi/redact/tiff.py
```diff
@@ -3,20 +3,28 @@
...
 
 from .redaction_plan import RedactionPlan
 
 if TYPE_CHECKING:
-    from tifftools.tifftools import IFD, TiffInfo
+    from tifftools.tifftools import IFD, TagEntry, TiffInfo
 
 
-class TiffMetadataRedactionPlan(RedactionPlan):
+class TiffRedactionPlan(RedactionPlan):
```
brianhelba, line 27: Can you split this trivial rename into a separate PR? It'd reduce the scope of this PR and make it easier to review?

imagedephi/redact/svs.py
```diff
@@ -82,6 +85,15 @@ def __init__(
                 else:
                     self.no_match_description_keys.add(key)
 
+    def get_associated_image_key_for_ifd(self, ifd: IFD):
+        """Attempt to match an associated image with a rule. Fallback to default rule."""
+        image_description = str(ifd["tags"][270]["data"])
```
brianhelba, line 90: `270` is magic. Is there any symbolic identifier for this tag?

imagedephi/redact/tiff.py
```diff
@@ -100,17 +147,83 @@ def report_missing_rules(self) -> None:
...
+        print(f"Found {len(self.image_redaction_steps)} associated images")
+        if len(self.image_redaction_steps):
+            default_rule = list(self.image_redaction_steps.values())[0]
+            print(f"Redaction action: {default_rule.action}")
+
+    def create_new_image(self, ifd: IFD, rule: ImageReplaceRule, temp_dir: Path) -> Path:
+        image = None
+        if rule.replace_with == "blank_image":
+            width = int(ifd["tags"][256]["data"][0])
+            length = int(ifd["tags"][257]["data"][0])
```
brianhelba, line 163: Are there symbolic constants for `256` and `257`?

stubs/tifftools/tifftools.pyi
manthey: `datapos` is not required for writing.   You could copy them when copying the tag (it will do no harm).  When combining ifds from two different files, tifftools has some bookkeeping about what file the data came from and the `StripOffsets` / `TileOffsets` (or other data blocks that have offsets and bytecount tags) use that to read the appropriate parts from the different source files.  If you are copying those tags to a new/different ifd, then you must also copy the `path_or_fobj` entry that is in the ifd so it knows where the data is being read from.

imagedephi/redact/tiff.py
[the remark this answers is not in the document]
naglepuff: In the case that `bytecounts` is present on a tag on the old IFD, what additional considerations need to be given when copying it over to the new IFD? Is what currently exists (copying datatype, count, and data) sufficient?

imagedephi/redact/svs.py
naglepuff: b536a25

imagedephi/redact/tiff.py
naglepuff: b536a25

imagedephi/redact/tiff.py
naglepuff: a49baef

imagedephi/redact/tiff.py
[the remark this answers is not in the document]
naglepuff: @manthey I've modified how I move data from the old IFD to to new IFD in dead62a. I'm using some [features of PIL](https://pillow.readthedocs.io/en/stable/reference/plugins.html#PIL.TiffImagePlugin.ImageFileDirectory_v2) to add the tags when saving the temporary image.

Let me know if this approach won't end up working for some reason, and I can revert the changes. Currently this only moves metadata of type ASCII over.

imagedephi/redact/tiff.py
brianhelba: Can you make this a PR, so we can merge it ASAP and get it out of the diff for this PR?

stubs/tifftools/tifftools.pyi
brianhelba: From a type annotation perspective, we should not make something `NotRequired` if an API guarantees that it'll be returned, as it forces us to implement unnecessary additional defensive code when using the API.

Maybe you need to split `TagEntry` into `ReadTagEntry` and `WriteTagEntry` classes, then annotate each part (places for reading vs writing) of the `tifftools` API with the appropriate object.

naglepuff: I'll revert this change since I'm trying a different approach to preserve metadata when creating a new image. If that approach doesn't end up working then I can look into splitting `TagEntry`.

imagedephi/redact/tiff.py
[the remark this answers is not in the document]
naglepuff: @manthey pinging for opinions on commit dead62a in particular

tests/test_redact.py
```diff
@@ -32,6 +32,15 @@ def test_redact_svs(svs_input_path, tmp_path, override_rule_set):
     assert b"ICC Profile" not in svs_output_file_bytes
 
 
+def test_redact_svs_image(data_dir, tmp_path):
```
brianhelba, line 35: Is this test necessary? It's just adding an additional assertion, but operating on effectively the same data as `test_redact_svs`. Could we make both redaction assertions in the existing test?

brianhelba (approved):

PR #103, opened 2023-05-12
Create redaction output file

marySalvi:
- Makes output_dir flag optional
- Removes overwrite flag
- Creates a new directory at the supplied output_dir named Redacted_[time_stamp] and places redacted images inside.
- Defaults to current directory if no output_dir is given.
- Uses freezegun for testing timestamped directories

Closes https://github.com/DigitalSlideArchive/ImageDePHI/issues/48 Closes https://github.com/DigitalSlideArchive/ImageDePHI/issues/72

### DIFF
--- a/imagedephi/redact/__init__.py
+++ b/imagedephi/redact/__init__.py
@@ -1,3 +1,3 @@
-from .redact import create_redact_dir, iter_image_files, redact_images, show_redaction_plan
+from .redact import iter_image_files, redact_images, show_redaction_plan
 
-__all__ = ["iter_image_files", "redact_images", "show_redaction_plan", "create_redact_dir"]
+__all__ = ["iter_image_files", "redact_images", "show_redaction_plan"]
--- a/imagedephi/redact/__init__.py
+++ b/imagedephi/redact/__init__.py
@@ -1,3 +1,3 @@
-from .redact import iter_image_files, redact_images, show_redaction_plan
+from .redact import create_redact_dir, iter_image_files, redact_images, show_redaction_plan
 
-__all__ = ["iter_image_files", "redact_images", "show_redaction_plan"]
+__all__ = ["iter_image_files", "redact_images", "show_redaction_plan", "create_redact_dir"]
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -53,7 +53,7 @@ def redact_images(
     base_rules = get_base_rules()
     images_to_redact = iter_image_files(input_path) if input_path.is_dir() else [input_path]
     time_stamp = datetime.datetime.now().isoformat(timespec="seconds")
-    redact_dir = Path(f"{output_dir}/Redacted_{time_stamp}")
+    redact_dir = output_dir / f"Redacted_{time_stamp}"
     try:
         redact_dir.mkdir(parents=True)
     except PermissionError:
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -40,13 +40,14 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
 
 
 def create_redact_dir(base_output_dir: Path) -> Path:
+    """Given a directory return a timestamped directory created in the base directory."""
     time_stamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
     redact_dir = base_output_dir / f"Redacted_{time_stamp}"
     try:
         redact_dir.mkdir(parents=True)
+        click.echo(f"Created redaction folder: {redact_dir}")
     except PermissionError:
-        "Cannnot create an output directory, permission error."
-    click.echo(f"Created redaction folder: {redact_dir}")
+        click.echo("Cannnot create an output directory, permission error.")
 
     return redact_dir
 
@@ -83,8 +84,13 @@ def redact_images(
             redaction_plan.report_missing_rules()
         else:
             redaction_plan.execute_plan()
-            output_path = _get_output_path(image_file, create_redact_dir(output_dir))
-            redaction_plan.save(output_path, overwrite)
+            try:
+                output_path = _get_output_path(image_file, create_redact_dir(output_dir))
+                redaction_plan.save(output_path, overwrite)
+            except FileNotFoundError:
+                click.echo(
+                    "Could not redact images, invalid output directory. Choose a writable directory"
+                )
 
 
 def show_redaction_plan(input_path: Path, override_rules: Ruleset | None = None) -> None:
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -41,7 +41,7 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
 
 
 def create_redact_dir(base_output_dir: Path) -> Path:
-    """Given a directory return a timestamped directory created in the base directory."""
+    """Given a directory, create and return a timestamped sub-directory within it."""
     time_stamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
     redact_dir = base_output_dir / f"Redacted_{time_stamp}"
     try:
@@ -65,9 +65,7 @@ def redact_images(
     try:
         redact_dir = create_redact_dir(output_dir)
     except PermissionError:
-        click.echo(
-            "Could not redact images, invalid output directory. Choose a writable directory"
-        )
+        click.echo("Could not redact images, invalid output directory. Choose a writable directory")
         sys.exit()
     for image_file in images_to_redact:
         if image_file.suffix not in FILE_EXTENSION_MAP:
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -44,6 +44,18 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
             yield child
 
 
+def create_redact_dir(base_output_dir: Path) -> Path:
+    time_stamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    redact_dir = base_output_dir / f"Redacted_{time_stamp}"
+    try:
+        redact_dir.mkdir(parents=True)
+    except PermissionError:
+        "Cannnot create an output directory, permission error."
+    click.echo(f"Created redaction folder: {redact_dir}")
+
+    return redact_dir
+
+
 def redact_images(
     input_path: Path,
     output_dir: Path,
@@ -52,13 +64,7 @@ def redact_images(
 ) -> None:
     base_rules = get_base_rules()
     images_to_redact = iter_image_files(input_path) if input_path.is_dir() else [input_path]
-    time_stamp = datetime.datetime.now().isoformat(timespec="seconds").replace(":", "")
-    redact_dir = output_dir / f"Redacted_{time_stamp}"
-    try:
-        redact_dir.mkdir(parents=True)
-    except PermissionError:
-        "Cannnot create an output directory, permission error."
-    click.echo(f"Created redaction folder: {redact_dir}")
+
     for image_file in images_to_redact:
         if image_file.suffix not in FILE_EXTENSION_MAP:
             click.echo(f"Image format for {image_file.name} not supported. Skipping...")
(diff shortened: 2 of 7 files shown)


marySalvi:
@brianhelba I am still having trouble with the windows tests on both branches. The tests are passing for macos and ubuntu on both branches though.

tox.ini
```diff
@@ -51,6 +51,7 @@ deps =
     pytest-asyncio
     pytest-mock
     pytest-timeout
+    freezegun
```
brianhelba, line 54: Make this alphabetical please.

imagedephi/redact/redact.py
```diff
@@ -54,8 +54,8 @@ def redact_images(
 ) -> None:
     base_rules = get_base_rules()
     images_to_redact = iter_image_files(input_path) if input_path.is_dir() else [input_path]
-    time_stamp = datetime.datetime.now().isoformat(timespec="seconds")
-    redact_dir = Path(f"{output_dir}/Redacted_{time_stamp}")
+    time_stamp = datetime.datetime.now().isoformat(timespec="seconds").replace(":", "")
```
brianhelba, line 57: This time format is still a bit hard for me to read. How about:

```suggestion
    time_stamp = datetime.datetime.now().strftime("%y-%m-%d-%H-%M-%S")
```

See:
* https://docs.python.org/3/library/datetime.html#datetime.datetime.strftime
* https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes

You could also decide to make the year be 4 digits. I don't have a strong preference.

This also should be more predictable if the `datetime.now()` has `microsecond` precision.

marySalvi: changed to this:
`Redacted_2023-05-15_15-58-58`
to me its easier to distinguish date from time with the underscore

imagedephi/redact/redact.py
```diff
@@ -54,8 +54,8 @@ def redact_images(
...
+    redact_dir = output_dir / f"Redacted_{time_stamp}"
```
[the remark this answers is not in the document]
marySalvi: update here 2bee621788f0c8f01498a18343ad5765d75fe00f
please double check that this is what you had in mind.

simple test for creation of directory added here:
e1ec843fec610bc0fb8fe16cf6147a8a77699ce7

tests/test_redact.py
```diff
@@ -22,12 +23,18 @@ def svs_input_path(data_dir, request) -> Path:
...
+
+
+@freeze_time("2023-05-12 12:12:53")
 def test_redact_svs(svs_input_path, tmp_path, override_rule_set):
     redact.redact_images(svs_input_path, tmp_path, override_rule_set)
 
-    svs_output_file = tmp_path / "REDACTED_test_svs_image_blank.svs"
-    assert svs_output_file.exists()
-    svs_output_file_bytes = svs_output_file.read_bytes()
+    output_file = tmp_path / "Redacted_2023-05-12_12-12-53/REDACTED_test_svs_image_blank.svs"
```
brianhelba, line 36: We cannot form paths this way on Windows. Each component following the `Path` must be its own string.

tests/test_e2e.py
```diff
@@ -29,13 +31,13 @@ def test_e2e_run(cli_runner: CliRunner, data_dir: Path, tmp_path: Path) -> None:
...
             str(data_dir / "input" / "tiff"),
+            "--output-dir",
             str(tmp_path),
         ],
     )
 
     assert result.exit_code == 0
-    output_file = tmp_path / "REDACTED_test_image.tif"
-    assert output_file.exists()
+    output_file = tmp_path / "Redacted_2023-05-12_12-12-53/REDACTED_test_image.tif"
```
brianhelba, line 40: These need to be separate components.

tests/test_e2e.py
```diff
@@ -95,8 +98,7 @@ def client_select_directory(port: int) -> httpx.Response:
 
     assert cli_result.exit_code == 0
     webbrowser_open_mock.assert_called_once()
-    output_file = tmp_path / "REDACTED_test_image.tif"
-    assert output_file.exists()
+    output_file = tmp_path / "Redacted_2023-05-12_12-12-53/REDACTED_test_image.tif"
```
brianhelba, line 101: These need to be separate components.

imagedephi/redact/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .redact import iter_image_files, redact_images, show_redaction_plan
+from .redact import create_redact_dir, iter_image_files, redact_images, show_redaction_plan
 
-__all__ = ["iter_image_files", "redact_images", "show_redaction_plan"]
+__all__ = ["iter_image_files", "redact_images", "show_redaction_plan", "create_redact_dir"]
```
brianhelba, line 3: `create_redact_dir` is basically an implementation detail of `imagedephi.redact.redact`. It doesn't need to be re-exported back to the `imagedephi.redact` module for general use.

If you need to access it for testing, you can always directly:
```import
from imagedephi.redact.redact import create_redact_dir
```

imagedephi/redact/redact.py
```diff
@@ -24,16 +25,7 @@ def _get_output_path(file_path: Path, output_dir: Path) -> Path:
     return output_dir / f"REDACTED_{file_path.name}"
 
 
-def _save_redacted_tiff(tiff_info: TiffInfo, output_path: Path, input_path: Path, overwrite: bool):
```
brianhelba, line 27: This is going to have conflicts with the changes in #95. Sorry. :disappointed:

imagedephi/redact/redact.py
```diff
@@ -55,11 +47,22 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
             yield child
 
 
+def create_redact_dir(base_output_dir: Path) -> Path:
+    time_stamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    redact_dir = base_output_dir / f"Redacted_{time_stamp}"
+    try:
+        redact_dir.mkdir(parents=True)
+    except PermissionError:
+        "Cannnot create an output directory, permission error."
```
brianhelba, line 56: What's this line doing? It's just a string.

tests/test_redact.py
```diff
@@ -22,12 +23,18 @@ def svs_input_path(data_dir, request) -> Path:
     return data_dir / "input" / request.param
 
 
+@freeze_time("2023-05-12 12:12:53")
+def test_create_redact_dir(tmp_path):
+    output_dir = redact.create_redact_dir(tmp_path)
+    assert output_dir.exists()
```
brianhelba, line 29: Since this is a unit test, I think it's fine to make stronger assertions about the behavior. I realize that the name is also implicitly tested by the other E2E tests, but it seems reasonable for a unit test to also assert that the name of `output_dir` matches what's expected.

tests/test_redact.py
```diff
@@ -22,12 +23,18 @@ def svs_input_path(data_dir, request) -> Path:
     return data_dir / "input" / request.param
 
 
+@freeze_time("2023-05-12 12:12:53")
+def test_create_redact_dir(tmp_path):
```
brianhelba, line 27: I don't necessarily think it's required now, but this unit test should hopefully make it easy to test that permission errors are handled (and knowing exactly what sort of filesystem states actually cause permission errors).

tests/test_redact.py
```diff
@@ -22,12 +23,18 @@ def svs_input_path(data_dir, request) -> Path:
...
+    output_dir = redact.create_redact_dir(tmp_path)
```
brianhelba, line 28: Perhaps you could make this a more challenging test case, by adding a non-existent additional level after `tmp_path`. If future changes remove `parents=True`, the test should fail.

imagedephi/redact/redact.py
```diff
@@ -55,11 +47,22 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
             yield child
 
 
+def create_redact_dir(base_output_dir: Path) -> Path:
```
brianhelba, line 50: Since this is a simple clean function, it might be nice to have a short docstring explaining its black-box behavior.

tests/test_redact.py
marySalvi: fixed in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

tests/test_e2e.py
marySalvi: fixed in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

tests/test_e2e.py
marySalvi: fixed in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

imagedephi/redact/__init__.py
marySalvi: fixed in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

imagedephi/redact/redact.py
marySalvi: I think i accidentally removed the click.echo :woman_facepalming:  fixed in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

tests/test_redact.py
marySalvi: added in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

tests/test_redact.py
marySalvi: added in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

imagedephi/redact/redact.py
marySalvi: added in a0c2bf04704b088fb3ba1cfdeb17ff930b8983bf

tests/test_redact.py
marySalvi: use this function to address #58 (in future PR)

imagedephi/redact/redact.py
```diff
@@ -37,6 +38,19 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
...
+    """Given a directory return a timestamped directory created in the base directory."""
+    time_stamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    redact_dir = base_output_dir / f"Redacted_{time_stamp}"
+    try:
+        redact_dir.mkdir(parents=True)
+        click.echo(f"Created redaction folder: {redact_dir}")
+    except PermissionError:
+        click.echo("Cannnot create an output directory, permission error.")
```
brianhelba, line 49: What's the intended control flow after this point? We're catching an exception and allowing `return redact_dir` to run normally afterwards, but we know the directory doesn't actually exist.

imagedephi/redact/redact.py
```diff
@@ -66,8 +80,13 @@ def redact_images(
             redaction_plan.report_missing_rules()
         else:
             redaction_plan.execute_plan()
-            output_path = _get_output_path(image_file, output_dir)
-            redaction_plan.save(output_path, overwrite)
+            try:
+                output_path = _get_output_path(image_file, create_redact_dir(output_dir))
+                redaction_plan.save(output_path, overwrite)
```
brianhelba, line 85: If possible, try to put only the lines which are expected to raise the exception inside the `try` block.

imagedephi/redact/redact.py
```diff
@@ -37,6 +38,19 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
             yield child
 
 
+def create_redact_dir(base_output_dir: Path) -> Path:
+    """Given a directory return a timestamped directory created in the base directory."""
+    time_stamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    redact_dir = base_output_dir / f"Redacted_{time_stamp}"
+    try:
+        redact_dir.mkdir(parents=True)
+        click.echo(f"Created redaction folder: {redact_dir}")
```
brianhelba, line 47: Since this `click.echo` isn't expected to fail, it shouldn't go inside the `try`.

Consider putting it inside an `else` if you want to keep it associated with the overall `try` block.

imagedephi/redact/redact.py
```diff
@@ -37,6 +38,19 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
             yield child
 
 
+def create_redact_dir(base_output_dir: Path) -> Path:
+    """Given a directory return a timestamped directory created in the base directory."""
```
brianhelba, line 42:
```suggestion
    """Given a directory, create and return a timestamped sub-directory within it."""
```

PR #107, opened 2023-05-16
Propagate HTTP endpoint exceptions to the CLI

brianhelba:
This will cause the CLI to exit with a non-zero status code on errors.

### DIFF
--- a/imagedephi/gui.py
+++ b/imagedephi/gui.py
@@ -24,8 +24,14 @@ def _load_template(template_name: str) -> str | None:
 async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     # Reset server state on startup, to support unit testing
     shutdown_event.clear()
+    app.state.last_exception = None
 
     yield
+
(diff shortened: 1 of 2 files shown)


imagedephi/gui.py
```diff
@@ -24,9 +24,16 @@ def _load_template(template_name: str) -> str | None:
 async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     # Reset server state on startup, to support unit testing
     shutdown_event.clear()
+    # TODO: Use ASGI lifespan state, once it's supported by Hypercorn
```
marySalvi, line 27: I think this would be better as an issue so we can track it and resolve when we are able.

marySalvi (changes requested):

imagedephi/gui.py
brianhelba: Done: #113

marySalvi (approved):

PR #111, opened 2023-05-16
Add a debug mode to improve GUI development

brianhelba:


brianhelba:
Maybe we should consider a "debug" mode and only mount this when that's enabled.

PR #120, opened 2023-05-26
Handle run failure

marySalvi:
Reports the redaction plan on run to show any images that will not be redacted and why.

### DIFF
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Generator
 import datetime
 import importlib.resources
+import os
 from pathlib import Path
 
 import click
@@ -36,14 +37,11 @@ def get_base_rules():
 
 
 def iter_image_files(directory: Path) -> Generator[Path, None, None]:
-    """
-    Given a directory return an iterable of available images.
-
-    May raise a PermissionError if the directory is not readable.
-    """
+    """Given a directory return an iterable of available images."""
     for child in directory.iterdir():
-        if child.is_file() and filetype.is_image(child):
-            yield child
+        if child.is_file() and os.access(child, os.R_OK):
+            if filetype.is_image(child):
+                yield child
 
 
 def create_redact_dir(base_output_dir: Path) -> Path:
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -6,6 +6,7 @@ import importlib.resources
 from pathlib import Path
 
 import click
+import filetype
 import tifftools
 import tifftools.constants
 import yaml
@@ -41,7 +42,7 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
     May raise a PermissionError if the directory is not readable.
     """
     for child in directory.iterdir():
-        if child.suffix in FILE_EXTENSION_MAP:
+        if child.is_file() and filetype.is_image(child):
             yield child
 
 
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -72,6 +72,7 @@ def redact_images(
     output_file_counter = 1
     output_file_max = len(images_to_redact)
     redact_dir = create_redact_dir(output_dir)
+    show_redaction_plan(input_path)
     for image_file in images_to_redact:
         if image_file.suffix not in FILE_EXTENSION_MAP:
             click.echo(f"Image format for {image_file.name} not supported. Skipping...")
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -74,36 +74,22 @@ def redact_images(
     redact_dir = create_redact_dir(output_dir)
     show_redaction_plan(input_path)
     for image_file in images_to_redact:
-        if image_file.suffix not in FILE_EXTENSION_MAP:
-            click.echo(f"Image format for {image_file.name} not supported. Skipping...")
-            continue
-        try:
-            redaction_plan = build_redaction_plan(image_file, base_rules, override_rules)
-        except tifftools.TifftoolsError:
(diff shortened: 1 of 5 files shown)


imagedephi/redact/redact.py
```diff
@@ -66,6 +67,7 @@ def redact_images(
     except PermissionError:
         click.echo("Could not redact images, invalid output directory. Choose a writable directory")
         sys.exit()
+    show_redaction_plan(input_path)
```
marySalvi, line 70: Scaled back to this simple solution in light of pending logger work. My original thought and attempt was to have a new prompt for each image that didn't have a plan, however knowing now that a user is likely to run in a batch I think simple reporting is the proper way to go.

As a follow up we should likely have a way to hand specific images instead of just a directory so that a user can feed in the images that are reported as not supported here. Filed corresponding issue: #121

imagedephi/redact/redact.py
```diff
@@ -35,8 +37,9 @@ def iter_image_files(directory: Path) -> Generator[Path, None, None]:
     May raise a PermissionError if the directory is not readable.
     """
     for child in directory.iterdir():
-        if child.suffix in FILE_EXTENSION_MAP:
-            yield child
+        if child.is_file() and os.access(child, os.R_OK):
+            if filetype.is_image(child):
+                yield child
```
marySalvi, line 42: @brianhelba Is this a good way to handle the permissions errors that `filetype` is getting with mac and windows or should this be handled elsewhere?

marySalvi: @brianhelba we had discussed this briefly but you had also mentioned possible better approaches. Let me know any suggestions you have.

imagedephi/redact/redact.py
```diff
@@ -59,31 +58,18 @@ def redact_images(
...
-            redaction_plan.save(output_path, overwrite)
+        if image_file.suffix in FILE_EXTENSION_MAP:
+            redaction_plan = build_redaction_plan(image_file, base_rules, override_rules)
+            if not redaction_plan.is_comprehensive():
+                click.echo(f"Redaction could not be performed for {image_file.name}.")
+                redaction_plan.report_missing_rules()
+            else:
+                redaction_plan.execute_plan()
+                output_path = _get_output_path(image_file, redact_dir)
+                redaction_plan.save(output_path, overwrite)
```
marySalvi, line 73: @naglepuff just wanted to point out this refactor. I removed these checks since we are now calling `show_redaction_plan` when `run` is called

naglepuff: I like having the checks in only one place. My only concern is that exceptions can still be raised in `build_redaction_plan`, so they should be handled in some other way here. Maybe there's a way to keep track of which files raised an exception in `show_redaction_plan`?

Once we have a more robust way of file type checking, I think we can get rid of a lot of these checks.

imagedephi/redact/redact.py
```diff
@@ -29,14 +31,11 @@ def get_base_rules():
...
-    Given a directory return an iterable of available images.
-
-    May raise a PermissionError if the directory is not readable.
-    """
+    """Given a directory return an iterable of available images."""
     for child in directory.iterdir():
-        if child.suffix in FILE_EXTENSION_MAP:
-            yield child
+        if child.is_file() and os.access(child, os.R_OK):
+            if filetype.is_image(child):
```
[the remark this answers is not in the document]
naglepuff: https://github.com/DigitalSlideArchive/ImageDePHI/issues/129

imagedephi/redact/redact.py
```diff
@@ -29,14 +31,11 @@ def get_base_rules():
...
-    """
-    Given a directory return an iterable of available images.
-
-    May raise a PermissionError if the directory is not readable.
-    """
+    """Given a directory return an iterable of available images."""
     for child in directory.iterdir():
-        if child.suffix in FILE_EXTENSION_MAP:
-            yield child
+        if child.is_file() and os.access(child, os.R_OK):
```
brianhelba, line 36: [Given](https://docs.python.org/3/library/os.html#os.access):
> Note I/O operations may fail even when [access()](https://docs.python.org/3/library/os.html#os.access) indicates that they would succeed, particularly for operations on network filesystems which may have permissions semantics beyond the usual POSIX permission-bit model.

I'm worried that a very common source of permission-denied errors (users accessing files from a network share) will not be fully handled.

imagedephi/redact/redact.py
```diff
@@ -35,14 +35,17 @@ def get_base_rules():
...
-            yield child
+        # Use first four bits to check if its a tiff file
+        if child.is_file():
+            try:
+                data = open(child, "rb").read(4)
+            except PermissionError:
+                pass
+            else:
+                if data in (b"II\x2a\x00", b"MM\x00\x2a", b"II\x2b\x00", b"MM\x00\x2b"):
+                    yield child
```
marySalvi, line 48: Although I know it is not best practice to catch an error and then do nothing with it I do think its appropriate in this instance. We want to produce a list of tiffs here so if we cannot read it I think it is appropriate to just not include it in the list.
@manthey @brianhelba

imagedephi/redact/redact.py
marySalvi: I _think_ these concerns are handled in a combination of the addition of using the magic bits to identify tiffs and also the pending logging PR. If I am wrong let me know and I'll be happy to find a suitable solution.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "hypercorn",
     "pyyaml",
     "Pillow",
+    "filetype",
```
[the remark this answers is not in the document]
marySalvi: removed in 9a70538f49cea37a4ac51621378248c3c0c15763

manthey (approved):

PR #122, opened 2023-05-26
Don't commit Redacted images

marySalvi:
Adding to gitignore so we don't accidentally commit images redacted while testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Redacted Images
+*/REDACTED_*
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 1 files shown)


manthey (approved):

.gitignore
```diff
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Redacted Images
```
brianhelba, line 131: This file is boilerplate. Could you put the additional custom rules at the top of the file, where it's more visible and obvious that we've added it?

Otherwise, LGTM.

PR #123, opened 2023-05-29
Add logger

marySalvi:
First pass adding a logger. Uses python core logging. 

**Notes**
- Currently all statements are either info or error level. Should consider if some redaction 'failures' would be better categorized as warnings 
- Should address different logging approaches. ie file in CWD, user designated file, within the directory of redacted images, and to the console.


imagedephi/main.py
```diff
@@ -31,6 +53,28 @@ class ImagedephiContext:
...
     CONTEXT_SETTINGS["help_option_names"].append("/?")
 
+# set default logger to warning level
+logger.remove()
+logger.add(sys.stderr, level="WARNING")
+
+
+def set_logging_level(v: int, q: bool, log_file=None):
+    log_output = log_file if log_file else sys.stderr
+    if q:
```
manthey, line 63: Rather than removing and adding a logger, can't we just change its level?  Removing and adding surprises me.

Also, in other code, I've done something like: `level=max(1, logging.WARNING - 10 * (args.verbose - args.silent)` (where args.verbose is the number of `-v` and args.silent is in the number of `-q`, and `logging.WARNING` is our default level).   This lets someone squelch errors, too, and handles someone adding `-vvvvv` just because they want the maximum verbosity and don't know how many `-v` are useful.

marySalvi: loguru doesn't have a way to directly update an existing logger:
https://loguru.readthedocs.io/en/stable/resources/recipes.html#changing-the-level-of-an-existing-handler
If one of the other methods is more preferable, I am happy to do that. We could also stick with the standard python logger if that makes more sense here.

manthey: It seems a curios choice for loguru.  I leave the choice of using it up to you.  I'd still use a simple formula for setting the new log level based on the number of `v` and `q` arguments rather than the `match` expression, since that still has the benefits of squelching errors or handling excessive `v`s.

manthey (approved):

PR #125, opened 2023-05-30
Don't attempt to redact ImageJ, NDPI files

### LINKED ISSUE #117: Don't attempt to process ndpi or ImageJ files
Currently we don't support redaction on ndpi or ImageJ files, since these formats aren't proper tiff files. We should skip redacting these files for now, instead of reporting them as unredactable due to insufficient rules.

naglepuff:
Fixes #117 

These images can technically be opened by `tifftools`, but can't be written back out as those formats post redaction. Therefore, we should skip redaction on these kinds of images, instead of trying to redact and reporting as a failed redaction.

### DIFF
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -13,6 +13,7 @@ from imagedephi.rules import Ruleset
 
 from .build_redaction_plan import FILE_EXTENSION_MAP, build_redaction_plan
 from .svs import MalformedAperioFileError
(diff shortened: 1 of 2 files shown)


manthey:
This looks fine to me.  I had considered that we could have rules that would fail if specific condition were met, so these would be part of the rule file, but we *can't* let someone override this, as we can't write ndpi or ImageJ files as they aren't actually tiffs.

PR #130, opened 2023-06-09
Use better name for redacted images

### LINKED ISSUE #60: Renaming output images
The initial approach to naming output images is to prepend "REDACTED_" to the input file name. There's no guarantee that input file names will be safe, so there should be a more secure approach.

It would still be useful for users of the software to be able to determine which input files informed which output files, so this should be taken into consideration as well. One idea we discussed was to incorporate the SHA hash of the input image into the metadata of the output image.

naglepuff:
This PR updates how we name redacted files. Instead of prepending `REDACTED_` to the input file names, prefix defined in the rules and a number, which increments with each slide redacted so far in the current run of Image DePHI.

It introduces a new property for `RuleSet`s: `output_file_name`. If `output_file_name` is `redacted_image`, then redacted images will follow the naming pattern `redacted_image_1.svs`, `redacted_image_2.svs`, etc.

This makes it harder to map back to the original file without comparing the images in the output and input directories. Future improvements could be including the SHA of the original image in the redacted image metadata.

Fix #60

### DIFF
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -17,8 +17,16 @@ from .svs import MalformedAperioFileError
 from .tiff import UnsupportedFileTypeError
 
 
-def _get_output_path(file_path: Path, output_dir: Path, base_name: str, count: int) -> Path:
-    return output_dir / f"{base_name}_{count}{file_path.suffix}"
+def _get_output_path(
+    file_path: Path,
+    output_dir: Path,
+    base_name: str,
+    count: int,
+    max: int,
+) -> Path:
+    number_length = len(str(max))
+    number = str(count).zfill(number_length)
+    return output_dir / f"{base_name}_{number}{file_path.suffix}"
 
 
 def get_base_rules():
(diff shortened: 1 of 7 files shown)


manthey:
I think users will want to be able to have structured output file names (e.g., specify `study_slide_####` and generate a 1-based sequence of new file names).

The uuid is still much better than keeping the original file names, and we could have a small utility to compute the hash of the WSI for comparison.

naglepuff:
> I think users will want to be able to have structured output file names (e.g., specify study_slide_#### and generate a 1-based sequence of new file names).

I think this should be simple enough to do here. We could allow specifying what the prefix is in the rule file (and maybe use `"study_slide_"` as the default.

naglepuff:
@manthey updated to pad the number portion of the output paths with zeroes for better sorting.

imagedephi/redact/redact.py
```diff
@@ -17,8 +17,16 @@
...
+def _get_output_path(
+    file_path: Path,
+    output_dir: Path,
+    base_name: str,
+    count: int,
+    max: int,
+) -> Path:
+    number_length = len(str(max))
+    number = str(count).zfill(number_length)
+    return output_dir / f"{base_name}_{number}{file_path.suffix}"
```
manthey, line 29: I think you can just do `return output_dir / f"{base_name}_{number:0{len(str(max))}d}{file_path.suffix}"`

manthey: And that `d` might not be needed since `number` is an integer.

naglepuff: Thanks. This has been added in bdb5707

manthey (approved):

PR #131, opened 2023-06-13
Position independent options

marySalvi:


marySalvi:
@manthey @brianhelba  An attempt to make global, position independent options for verbosity. 
For the record, this was really fun.

PR #140, opened 2023-07-17
Zip up packaged executables

zachmullen:


README.md
```diff
@@ -6,18 +6,11 @@ ImageDePHI is an application to redact personal data (PHI) from whole slide imag
...
 
-* On **Linux and macOS only**, [make the file executable](https://askubuntu.com/a/229592):
-    ```bash
-    chmod +x ./Downloads/imagedephi-Linux
-    ```
-    or
-    ```bash
-    chmod +x ./Downloads/imagedephi-macOS
-    ```
+* Unzip the downloaded file, which will extract the executable named `imagedephi` (or `imagedephi.exe` on Windows).
```
marySalvi, line 9: This is great, it fixes an unexpected behavior on windows: the terminal flashing open and immediately closing with the command `start imagedephi.exe`
I do however think that the user still needs to make it executable for linux and mac though.

zachmullen: > I do however think that the user still needs to make it executable for linux and mac though.

Have you verified this? It should no longer be necessary because the zip file should preserve the file permissions.

.github/zip_and_upload_package.sh
```diff
@@ -0,0 +1,28 @@
...
+
+cd dist/
+
+if [ -f "./imagedephi.exe" ]; then
+    executable="imagedephi.exe"
+else
+    executable="imagedephi"
+fi
+
+chmod +x $executable
```
marySalvi, line 16: I do see it being made executable here, but testing locally I had to change it manually.

zachmullen: How did you test locally? I can try to reproduce on my end.

marySalvi: I downloaded the linux artifact here https://github.com/DigitalSlideArchive/ImageDePHI/actions/runs/5578038631
extracted the file then tried to run the program from the terminal. I got a permission error.

marySalvi: Ah I see! Thanks for explaining.

marySalvi: Just to close the loop on this. I tested again locally using https://github.com/nektos/act and confirmed the unzipped file is executable.

manthey (approved):

PR #143, opened 2023-08-02
Document rules format

naglepuff:
Add documentation to explain how ImageDePHI rules are formatted and how they are used during runs of the program.

Reviewers: please leave feedback regarding how the new documentation can be made more clear, and/or additional information that would be useful for end users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,3 +21,141 @@ From a command line, execute the application to get full usage help.
 Alternatively **on Windows only**, directly open `imagdephi-Windows.exe` in Windows Explorer to launch the ImageDePHI GUI.
 
 If running on macOS, you may need to [add the executable to the list of trusted software](https://support.apple.com/guide/mac-help/apple-cant-check-app-for-malicious-software-mchleab3a043/mac) to launch ImageDePHI in the same way you would any other registered app.
+
+# Rules
+Image redaction is determined by a set of rules. Each run of the program can use a different set of rules, depending on how the command was run. By default, the base set of rules are used. These rules are provided by the `imagedephi` package and can be found [here](https://github.com/DigitalSlideArchive/ImageDePHI/blob/main/imagedephi/base_rules.yaml).
+
+## Rule Application
+All runs of `imagedephi` use the provided base set of rules as a foundation. End users can use the ruleset framework to buld custom rulesets that handle additional or custom metadata not covered by the base rules, or override the behavior of the base rule set.
+
+These override rule sets can be specified by using the `-r my_ruleset.yml` or `--override-rules my_ruleset.yml` flag. This flag is available for both the `imagedephi run` and `imagedephi plan` subcommands.
+
+When `imagedephi` determines the steps to redact a file, it checks each piece of metadata in the file. For each piece of metadata found this way, it will first consult the override rule set, if present, for an applicable rule. If the override rule set does not contain a rule for that piece of metadata, the program will check the base ruleset.
+
+If neither the override rule set or base rule set cover a piece of metadata, redaction will fail, and the program will list the metadata that it could not redact. There is no default behavior for unknown metadata.
+
+
+## Ruleset Format Overview
+In order to read the base rules and build your own custom rule sets, it is important to understand the format in which rulesets are specified. Rulesets are defined by `.yml` files (one ruleset per file), and are a dictionary with the following top-level tags: `name`, `description`, `output_file_name`, `tiff`, and `svs`.
+
+### Generic Properties
+The following three properties belong to the rulesets themselve, and don't influence redaction behavior. Redaction behavior is specified by file type.
+
+#### `name`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,3 +21,141 @@ From a command line, execute the application to get full usage help.
...
+    metadata:
+```
+have `metadata_keys` for particular tiff tags (e.g. `ImageDescription`, `ImageWidth`).
+
+Available actions for metadata rules are:
+
+* `delete`: the metadata will not appear in the output file
+* `keep`: the metadata will appear unchanged in the output file
+* `replace`: replace the metadata with a specified value. If this is the `action`, additional fields are required.
+* `check_type`: the metadata will appear in the output file if it is of the expected type. Requires additional fields
```
manthey, line 121: Am I correct that the rule matches if the type matches but not if it doesn't?  That is, a match is `keep` but a failure to match is a rule miss, so you would add another rule to replace or delete those non-matching type entries?

README.md
```diff
@@ -21,3 +21,141 @@ From a command line, execute the application to get full usage help.
...
+svs:
+    associated_images:
+        ...
+    metadata:
+        ...
+    image_description:
+        ...
+```
+
+The keys for the `metadata` rules are the names of tiff tags defined by the tiff standard.
```
manthey, line 159: In tifftools there are some variations of tag names to try to be inclusive to what other libraries do.  Are we only using the primary name or allowing any match?  I'd also comment on case sensitivity (or lack thereof) here.

README.md
naglepuff: The initial implementation of the type check rules matches by tag/image description key. Behavior is branching based on the if the actual datatype lines up with the expected data type, but both are considered a "match."

I can see why this might be confusing. I think that if a type mismatch was a rule "miss," and another rule was required to fully redact some piece of metadata, it might be unintuitive. With the initial implementation, only one rule is needed to redact that piece of metadata. I'm open to changing how that type of rule works if the current implementation is lacking in some way.

README.md
naglepuff: There's a [utility function ](https://github.com/DigitalSlideArchive/ImageDePHI/blob/46989ccde270bf725ad6fb6f8948fdc4e7ef7b65/imagedephi/utils/tiff.py#L4) that we use to get tiff tags. I think it satisfies allowing any match:

```python
>>> from imagedephi.utils.tiff import get_tiff_tag
>>> get_tiff_tag("GrayResponseUnit")
290
>>> get_tiff_tag("GreyResponseUnit")
290
```

README.md
```diff
@@ -21,3 +21,141 @@ From a command line, execute the application to get full usage help.
...
+When `imagedephi` determines the steps to redact a file, it checks each piece of metadata in the file. For each piece of metadata found this way, it will first consult the override rule set, if present, for an applicable rule. If the override rule set does not contain a rule for that piece of metadata, the program will check the base ruleset.
+
+If neither the override rule set or base rule set cover a piece of metadata, redaction will fail, and the program will list the metadata that it could not redact. There is no default behavior for unknown metadata.
+
+
+## Ruleset Format Overview
+In order to read the base rules and build your own custom rule sets, it is important to understand the format in which rulesets are specified. Rulesets are defined by `.yml` files (one ruleset per file), and are a dictionary with the following top-level tags: `name`, `description`, `output_file_name`, `tiff`, and `svs`.
+
+### Generic Properties
+The following three properties belong to the rulesets themselve, and don't influence redaction behavior.
```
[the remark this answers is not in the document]
naglepuff: Fixed typo in 1986829

README.md
```diff
@@ -21,3 +21,141 @@ From a command line, execute the application to get full usage help.
...
+
+#### `name`
+Provide a name for a ruleset. This is used by the `imagedephi plan` command to specify which ruleset is being used to redact a particular piece of metadata.
+
+#### `description`
+You can add a description to your custom rulesets. This is not used by the program, but can be helpful to communicate what cases your custom rulesets are designed for.
+
+#### `output_file_name`
+Specify how the output files should be names here. The base ruleset contains the value `study_slide`. In this case, if my input slides are named: `john_smith_lung.svs` and `john_smith_pancreas.svs`, the redacted output images will be named `study_slide_1.svs` and `study_slide_2.svs`.
```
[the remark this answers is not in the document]
naglepuff: Fixed in 1986829

README.md
manthey: I think this means we need to clarify the text: `This will either keep the metadata if the type matches or delete the metadata if the type does not match`.  And, looking at the code, if we are using a type check of rational, should count be the number of rationals or the number of integers (that is, count of 1 would be two integers)?

README.md
manthey: Great.  Let's say that `Names are case insensitive and common variations are accepted.`

naglepuff: Done in cec9d21

manthey (approved):

PR #149, opened 2023-08-08
And endpoint to retrieve associated images and thumbnails

### LINKED ISSUE #138: Explore showing macro, label images in UI

naglepuff:
Fix #138 

Caveats: 

1. Has a limited way to look for associated images. It checks `ImageDescription`, which can find `label` and `macro` images for svs files, but not always. It falls back to `NewSubfileType`, which can find `macro` images
2. Doesn't try very hard to extract a thumbnail. If the tiled images are compressed with a method other than JPEG, it doesn't work

UI integration is for demonstration/testing, and will likely not make it through the review process for this PR, as a UI overhaul is in progress, and the new endpoint can be integrated in to that as part of a separate PR.

### DIFF
--- a/imagedephi/templates/BaseSelectorWidget.html.j2
+++ b/imagedephi/templates/BaseSelectorWidget.html.j2
@@ -36,6 +36,8 @@
     <li>
       <i class="mdi mdi-image text-sky-800"></i>
       {{ child_image.name }}
+      <img src="{{ label_image_url(child_image.name) }}" alt="label image"/>
+      <img src="{{ macro_image_url(child_image.name) }}" alt="macro image"/>
     </li>
   {% endfor %}
   {% if directory_data.child_images|length > 10 %}
--- a/imagedephi/templates/BaseSelectorWidget.html.j2
+++ b/imagedephi/templates/BaseSelectorWidget.html.j2
@@ -34,10 +34,12 @@
 <ul>
   {% for child_image in directory_data.child_images[:10] %}
     <li>
-      <i class="mdi mdi-image text-sky-800"></i>
-      {{ child_image.name }}
-      <img src="{{ label_image_url(child_image.name) }}" alt="label image"/>
-      <img src="{{ macro_image_url(child_image.name) }}" alt="macro image"/>
+      <div class="inline-flex">
+        <img src="{{ label_image_url(child_image.name) }}" alt="label image"/>
+        <img src="{{ macro_image_url(child_image.name) }}" alt="macro image"/>
+        <i class="mdi mdi-image text-sky-800"></i>
+        {{ child_image.name }}
+      </div>
     </li>
   {% endfor %}
   {% if directory_data.child_images|length > 10 %}
--- a/imagedephi/templates/BaseSelectorWidget.html.j2
+++ b/imagedephi/templates/BaseSelectorWidget.html.j2
@@ -35,8 +35,9 @@
   {% for child_image in directory_data.child_images[:10] %}
     <li>
       <div class="inline-flex">
-        <img src="{{ label_image_url(child_image.name) }}" alt="label image"/>
-        <img src="{{ macro_image_url(child_image.name) }}" alt="macro image"/>
+        <img src="{{ image_url(child_image.name, 'label') }}" alt="label image"/>
+        <img src="{{ image_url(child_image.name, 'macro') }}" alt="macro image"/>
+        <img src="{{ image_url(child_image.name, 'thumbnail') }}" alt="thumbnail image"/>
         <i class="mdi mdi-image text-sky-800"></i>
         {{ child_image.name }}
       </div>
--- a/imagedephi/templates/BaseSelectorWidget.html.j2
+++ b/imagedephi/templates/BaseSelectorWidget.html.j2
@@ -34,14 +34,8 @@
 <ul>
   {% for child_image in directory_data.child_images[:10] %}
     <li>
-      <div class="inline-flex">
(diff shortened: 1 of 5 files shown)


imagedephi/templates/BaseSelectorWidget.html.j2
```diff
@@ -34,8 +34,14 @@
...
-      <i class="mdi mdi-image text-sky-800"></i>
-      {{ child_image.name }}
+      <div class="inline-flex">
+        <img src="{{ image_url(child_image.name, 'label') }}" alt="label image" />
+        <img src="{{ image_url(child_image.name, 'macro') }}" alt="macro image" />
+        <img src="{{ image_url(child_image.name, 'thumbnail') }}"
+             alt="thumbnail image" />
+        <i class="mdi mdi-image text-sky-800"></i>
+        {{ child_image.name }}
+      </div>
```
naglepuff, line 44: I expect to remove this before the PR is merged, but this could be useful for reviewers testing the current GUI using the new image endpoints.

imagedephi/utils/tiff.py
```diff
@@ -1,5 +1,33 @@
...
+                tag_id,
+                tagSet=tag_set,
+                datatype=tifftools.Datatype[entry["datatype"]],
+            )
+            if tag.isIFD():
+                # entry['ifds'] contains a list of lists
+                # see tifftools.read_tiff
+                for sub_ifds in entry.get("ifds", []):
+                    yield from iter_ifds(sub_ifds, tag.get("tagset"))
+        yield ifd
```
naglepuff, line 29: It looks like an IFD iterator would be useful outside of the `RedactionPlan` classes. I think refactoring could be done in a follow up PR to keep the scope of this one down.

imagedephi/utils/tiff.py
```diff
@@ -13,3 +41,59 @@ def get_tiff_tag(tag_name: str) -> tifftools.TiffTag:
...
+
+    min_width = float("inf")
+    lowest_res_ifd = None
+    for ifd in iter_ifds(image_info["ifds"]):
+        # We are interested in the lowest res tiled image.
+        if tifftools.Tag.TileWidth.value not in ifd["tags"]:
+            continue
+        # PIL can only read JPEG-compressed tiff images
+        if ifd["tags"][tifftools.Tag.Compression.value] != tifftools.constants.Compression.JPEG:
+            continue
```
naglepuff, line 84: @manthey Are there other compression formats that we might come across that PIL would support?

imagedephi/utils/tiff.py
```diff
@@ -13,3 +41,59 @@ def get_tiff_tag(tag_name: str) -> tifftools.TiffTag:
...
+
+    return lowest_res_ifd
+
+
+def get_is_svs(image_path: Path) -> bool:
+    image_info = tifftools.read_tiff(image_path)
+    if tifftools.Tag.ImageDescription.value not in image_info["ifds"][0]:
+        return False
+    image_description = image_info["ifds"][0]["tags"][tifftools.Tag.ImageDescription.value]["data"]
+    return "aperio" in str(image_description).lower()
```
naglepuff, line 99: @manthey We've recently seen svs images where associated image IFDs don't have an image description. Is the first IFD always going to be a tiled image? This function could just iterate over all IFDs and return `True` if *any* `ImageDescription` contains `"Aperio"`

imagedephi/templates/BaseSelectorWidget.html.j2
manthey: This was handy.  It shows that on TCGA-5N-A9KM-01Z-00-DX1.5197F750-D17F-459B-B74D-846F5F50F7B7.svs I'd expect to get a label image, but don't.  Overall, this PR looks good.

manthey: You can find a copy of the mentioned file on my local server under /collection/TCGA/square

naglepuff: I'm a little bit confused about this image. There is indeed a label image, (Directory 5 when looking at the output of `tifftools dump`) but it doesn't have an `ImageDescription` tag pointing that out. 

The value of `NewSubfileType` for this directory is `1`, but that seems odd. According to tifftools, this bit represents "Image is a reduced-resolution version of another image in this TIFF file," but the label image isn't a reduced resolution version of any image in the file (that I'm aware of).

Additionally, other directories in the image, apart from the label and macro images, have `NewSubfileType = 0`. I would expect all but the highest resolution image to hava a value of `1`. What's going on here? Is there another heuristic for finding a label image among the IFDs? If large image does this well, could you point me in that direction?

manthey:
I've mirrored our logic for determining label/macro images in https://github.com/girder/large_image/pull/1291.

manthey:
This is looking good for what is here.  @naglepuff Do you want to remove the UI integration and then merge this?  Or should we create a fall-back reader for a tiled layer that doesn't decode with PIL-as-a-tiff first?

naglepuff:
> This is looking good for what is here. @naglepuff Do you want to remove the UI integration and then merge this? Or should we create a fall-back reader for a tiled layer that doesn't decode with PIL-as-a-tiff first?

I'll take out the UI elements and merge in the rest.

imagedephi/templates/BaseSelectorWidget.html.j2
naglepuff: Based on our conversation on Tuesday, I updated the logic in ad0302d

manthey (approved):

PR #150, opened 2023-08-14
Demonstrate removal of orphaned/old metadata

### LINKED ISSUE #146: Test redacting metadata with no pointer
Using tools like `libtiff` to set metadata can result in a tiff file where old metadata is still around, but not referenced by a tag. This can be an issue if a user uses `tiffset` to change metadata that contained PHI.

We should add a test to ImageDePHI that demonstrates that we would sanitize such data. 

In order to effectively test we would need to use `tiffset` on a file to add a new piece of metadata to an image, confirm that the old value and new value are both contained somewhere in the file, and then show that only the new value is in the file after being run through ImageDePHI.

naglepuff:
Fix #146 

Add test image and test to show that if stale metadata exists in a tiff file, our software redacts that. This is useful to show since tools like `tiffset` will add/change metadata without deleting the old bytes.


imagedephi/redact/redact.py
```diff
@@ -36,7 +36,7 @@ def get_base_rules():
 
 def iter_image_files(directory: Path) -> Generator[Path, None, None]:
     """Given a directory return an iterable of available images."""
-    for child in directory.iterdir():
+    for child in sorted(directory.iterdir()):
```
naglepuff, line 39: @manthey assuming that tests are now passing on this branch, then the issue was sort order for iterating over files in the directory.

This doesn't seem like a great solution to me, since we're now loading the entire list of files into memory, and a use case would be redacting a ton of images at a time, so this seems wasteful.

Would you have a better idea for fixing the failing test without having to add this inefficiency?

I think that if ImageDePHI did have a way to link a redacted file to an input file, we could utilize that during testing to check the correct file.

Additionally, I think rearranging the test files so that the newly added image was in its own subdirectory would be better than using `sorted()` here. Does that make sense to you?

manthey: I see the point about sorting a large number of files.

For now, I think rearranging the test files OR just redact specific files instead of folders is the right approach

I wrote a crude first pass at generating some hashing for the contents of a tiff (see https://gist.github.com/manthey/991caf81dee52c14f2b8e1d463356baf).  For images without z-stacks/multi-channels, etc., comparing the (0,) hash entry from that gist would be sufficient.  But I don't think we want to do that for most tests, as it is a comparatively expensive operation -- maybe as part of a different test, especially if we polish that gist into something more generally usable.

naglepuff: I moved the new test file out of the `data/input/tiff` directory. Looks like all of the tests are passing now. Let me know if there's anything else I need to do here.

manthey (approved):

PR #154, opened 2023-09-13
Extract thumbnail image data when PIL can't read an IFD

### LINKED ISSUE #151: Extract thumbnails when PIL can't decode
Sometimes, PIL will be unable to decode an IFD. This can cause problems when we try to extract thumbnail images for the GUI.

There are 2 reasons why PIL will fail to read an IFD. Either the image is too big, or uses a compression format that the PIL tiff reader can't decode.

For now, we won't worry about images being too big, but instead we will focus on compression formats that the PIL tiff reader can't decode.

Once we find the associated image IFD we want to use for the thumbnail, we will have to use the tile or stripe offsets to extract a chunk of the image at a time, and rebuild the image for PIL to read just the image data.

naglepuff:
Fix #151 

This PR adds a fallback method to obtain a thumbnail image from a tiff/svs file when PIL can't open a thumbnail IFD. This is necessary because PIL's tiff reader can't read tiff images that are JPEG2k encoded.

In this case, the low res image determined to be the thumbail is extracted from the original tiff/svs image data.

To see the endpoint in action, run `imagedephi gui` and point your browser towards: 
`localhost:<port>/image/?file_name=<path/to/file>&image_key=thumbnail`

### DIFF
--- a/imagedephi/gui.py
+++ b/imagedephi/gui.py
@@ -125,7 +125,7 @@ def select_directory(
     )
 
 
-def get_composite_image(ifd: IFD, file_name: str):
+def extract_thumbnail_from_image_bytes(ifd: IFD, file_name: str):
     offsets = ifd["tags"][tifftools.Tag.TileOffsets.value]["data"]
     byte_counts = ifd["tags"][tifftools.Tag.TileByteCounts.value]["data"]
     num_tiles = len(offsets)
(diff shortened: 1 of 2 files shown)


imagedephi/gui.py
```diff
@@ -124,7 +125,54 @@ def image_url(path: str, key: str) -> str:
...
+            bottom = top + tile_size[0]
+            right = left + tile_size[1]
+            if bottom > image_array.shape[0]:
+                bottom = image_array.shape[0]
+            if right > image_array.shape[1]:
+                right = image_array.shape[1]
+
+            x_width = bottom - top
+            y_length = right - left
+            image_array[top:bottom, left:right, :] = tile_array[0:x_width, 0:y_length, :]
```
manthey, line 157: If this is the only thing we are using numpy for, we could have done this entirely with PIL.  But if numpy bundles properly for our executables, we should leave this as written.

naglepuff: I was able to get rid of `numpy` in favor of `Image.paste`. See e28ffa6 for the changes.

manthey (approved):

manthey:
Nothing to do with this PR directly, but I noticed it while testing.  We have a constant saying how tall an image would be in the GUI but not the maximum width.  When we hook it up, we probably want to specify the maximum width and height (and maybe make them something the url specifies if the UI ever needs different sizes).

naglepuff:
> Nothing to do with this PR directly, but I noticed it while testing. We have a constant saying how tall an image would be in the GUI but not the maximum width. When we hook it up, we probably want to specify the maximum width and height (and maybe make them something the url specifies if the UI ever needs different sizes).

#155

PR #164, opened 2023-10-04
UI redaction

marySalvi:
PR #161 should be merged first


marySalvi:
This currently hangs forever and I am stumped.  To test have to windows open one to `localhost:8000` the other to `localhost:8000/websocket`. Ideally when you make a selection for redaction the progress would be logged to the websocket endpoint. It does not.

manthey:
I see this says "Redacting image out of 25" but there is no count in that message and it doesn't update.

marySalvi:
@manthey This should be fixed in fae0172cbae16ef0af8f20af1e773fe178cb7aef. The url for the websocket had been hard coded which breaks the cli.

manthey (approved):
Approved as a quick and messy fix.

PR #173, opened 2023-12-01
Vue frontend

marySalvi:
This achieves feature parity with the current UI using vue instead of jinja templates.


manthey (approved):
This is working for me.  I checked the component PRs as they came along; I'd say merge this and we'll sort any issues I missed as we do UI work.

PR #174, opened 2023-12-05
Implement DICOM metadata redaction

naglepuff:
Here is the initial set of changes to support redacting DICOM images.

As a start, only metadata redaction is supported and only redaction of single .dcm files is supported.

(TODO before ready for full review/merge):

- [x] Add functional test
- [x] Clean up code (prints, misc. comments, etc

### DIFF
--- a/imagedephi/redact/dicom.py
+++ b/imagedephi/redact/dicom.py
@@ -1,0 +1,183 @@
+from __future__ import annotations
+
+from collections.abc import Generator
+from pathlib import Path
+from typing import TYPE_CHECKING
+from uuid import uuid4
+
+import pydicom
+from pydicom import valuerep
+from pydicom.datadict import keyword_for_tag
+from pydicom.dataelem import DataElement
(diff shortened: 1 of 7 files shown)


imagedephi/redact/dicom.py
```diff
@@ -0,0 +1,182 @@
...
+    metadata_redaction_steps: dict[int, ConcreteMetadataRule]
+    no_match_tags: list[BaseTag]
+    uid_map: dict[str, str]
+
+    @staticmethod
+    def _iter_dicom_elements(
+        dicom_dataset: Dataset,
+    ) -> Generator[tuple[DataElement, Dataset], None, None]:
+        for element in dicom_dataset:
+            if element.VR == "SQ":
```
manthey, line 69: Should we use the `valuerep.VR.SQ` enum rather than `"SQ"`?

manthey (approved):
I made one recommendation, but this works for me.  We can add more rules based on other sample files in a separate PR.

PR #175, opened 2023-12-13
Add flag to skip renaming files

### LINKED ISSUE #171: Specifying target file name
It seems that the app currently uses the parameter study_slide to specify the output file name, but I can't see anywhere that can be specified.

I think currently just generating a study_slide and incrementing it based on the order an image happens to be in a folder should not be the default behavior.  In most use cases, the user would either provide a very simple CSV file with. <CurrentName>,<NewName> or the app simply preserves the existing filename, but maybe adds. ".deid." which is what I have been doing internally during testing.

In most use cases, the user would not only hand over the raw images, but some basic manifest that maps the files to some basic metadata.  With the current implementation, there doesn't seem to be a simple way to even figure out the map of OldFileName --> NewFileName without doing a lot of work.

At the very least, exporting a deidManifest file with the mapping would allow the user to then update/modify them during whatever QC process they are going to run to make sure all the files they wanted to be processed were actually processed.

naglepuff:
Fix #171 

Add an option to the `run` command to keep original file names, in the case where users are confident that their image file names do not contain PHI but would be helpful to preserve after redaction.

To use: 

`imagedephi run path/to/images -o output/path --skip-rename`

### DIFF
--- a/imagedephi/gui.py
+++ b/imagedephi/gui.py
@@ -63,7 +63,7 @@ app = FastAPI(
 templates = Jinja2Templates(
     # Jinja2Templates requires a "directory" argument, but it is effectively unused
     # if a custom loader is passed
-    directory="",
+    directory=".",
     loader=FunctionLoader(_load_template),
 )
 
--- a/imagedephi/main.py
+++ b/imagedephi/main.py
@@ -111,12 +111,21 @@ def imagedephi(
     help="Path where output directory will be created.",
     type=click.Path(exists=True, file_okay=False, readable=True, writable=True, path_type=Path),
 )
+@click.option("--rename/--skip-rename", default=True)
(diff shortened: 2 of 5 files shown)


manthey (approved):

manthey:
Because of an upstream package change, you'll need #177 to allow CI to pass again.

naglepuff:
@manthey I merged branch `harden-some-tests` (the source for #177) into this branch to fix CI. Not sure if you wanted to merge #177 before this PR.

manthey:
> @manthey I merged branch `harden-some-tests` (the source for #177) into this branch to fix CI. Not sure if you wanted to merge #177 before this PR.

I'll have you approve #177 and then I'll merge it.

PR #176, opened 2023-12-28
Vue redact components

marySalvi:
This PR creates feature parity with the current frontend and utilizes vue instead of jinja:

**Functionality should be the same for cli and debug mode**
Homepage available on: `localhost:port:`

- Converts input and output directory selection modals to vue components
- Updates endpoints to provide json response instead of templates
- Adds progress bar for redaction
- Removes jinja templates and djlint
- Adds linting and formatting for frontend(eslint)

[Screencast from 12-28-2023 09:45:21 AM.webm](https://github.com/DigitalSlideArchive/ImageDePHI/assets/40494088/7b348cee-05b5-412e-bbc4-c891b1a207e7)

**TO DO**

- [x] - Address binary distribution. : addressed with #180  
- [x] - Update tests : addressed with another branch


manthey:
This doesn't work for me unless I make changes, such as 
```
diff --git a/imagedephi/gui.py b/imagedephi/gui.py
index 09499be..0aef6fe 100644
--- a/imagedephi/gui.py
+++ b/imagedephi/gui.py
@@ -68,22 +68,6 @@ if debug_mode:
         allow_headers=["*"],
     )
 
-else:
-    app.mount(
-        "/",
-        StaticFiles(
-            directory=str(importlib.resources.files("imagedephi") / "web_static"), html=True
-        ),
-        name="home",
-    )
-    app.mount(
-        "/assets",
-        StaticFiles(
-            directory=str(importlib.resources.files("imagedephi") / "web_static" / "assets")
-        ),
-        name="assets",
-    )
-
 
 shutdown_event = asyncio.Event()
 
@@ -207,7 +191,7 @@ def on_internal_error(request: Request, exc: Exception) -> PlainTextResponse:
     )
 
 
-@app.get("/directory/")
+@app.get("/directory")
 def select_directory(
     directory: str = "/",  # noqa: B008
 ):
@@ -260,7 +244,7 @@ def get_associated_image(file_name: str = "", image_key: str = ""):
     return get_image_response_from_ifd(ifd, file_name)
 
 
-@app.post("/redact/")
+@app.post("/redact")
 def redact(
     input_directory: str,  # noqa: B008
     output_directory: str,  # noqa: B008
@@ -290,3 +274,15 @@ async def websocket_endpoint(websocket: WebSocket):
             await websocket.send_json(message_dict)
         else:
             await asyncio.sleep(0.001)
+
+
+app.mount(
+    "/assets",
+    StaticFiles(directory=str(importlib.resources.files("imagedephi") / "web_static" / "assets")),
+    name="assets",
+)
+app.mount(
+    "/",
+    StaticFiles(directory=str(importlib.resources.files("imagedephi") / "web_static"), html=True),
+    name="home",
+)
```
then it works except for progress (the progress dialog never advances or goes away).

imagedephi/gui/utils/image.py
```diff
@@ -0,0 +1,91 @@
...
+from fastapi.responses import StreamingResponse
+import tifftools
+
+from imagedephi.gui.utils.constants import MAX_ASSOCIATED_IMAGE_HEIGHT
+
+if TYPE_CHECKING:
+    from tifftools.tifftools import IFD
+
+
+def extract_thumbnail_from_image_bytes(ifd, file_name: str) -> Image.Image | None:
```
marySalvi, line 14: this should be:
`def extract_thumbnail_from_image_bytes(ifd: IFD, file_name: str) -> Image.Image | None:`

but having the `IFD` type hint throws a NameError(?!)

Any insight into how to resolve this would be really appreciated. I can't figure out why python cares about the typehint only here.

@manthey @naglepuff

marySalvi:
> This is working for me in testing, but when the redaction is finished, it throws an error `AttributeError: 'FastAPI' object has no attribute 'shutdown_event`.

Fixed in 1e1f529701ae50b7891a988bef7402deec68fbb4

marySalvi:
> then it works except for progress (the progress dialog never advances or goes away).

Fixed in 55a74c5d2495ba79e940c36ef15e846d2b583e4d

Although it should be noted that the dialog does not go away on completion; this PR just matches the main branch functionality.

imagedephi/gui/api/api.py
```diff
@@ -0,0 +1,105 @@
...
+    output_path = Path(output_directory)
+    if not input_path.is_dir():
+        raise HTTPException(status_code=404, detail="Input directory not found")
+    if not output_path.is_dir():
+        raise HTTPException(status_code=404, detail="Output directory not found")
+
+    redact_images(input_path, output_path)
+
+    # Shutdown after the response is sent, as this is the terminal endpoint
+    background_tasks.add_task(shutdown_event.set)  # type: ignore[attr-defined]
```
marySalvi, line 92: What are our thoughts on removing the shutdown here? I don't know if this still makes sense with our workflow going forward and it causes a race condition with the progress component. 
I think with pivoting to an SPA that users will be able to continue to interact with the app after the redaction and this will no longer be a terminal endpoint.

I would appreciate any thoughts and feedback.
@manthey @brianhelba

manthey (approved):

PR #178, opened 2024-01-04
Allow redacting on subdirs

### LINKED ISSUE #170: Add Recursive option when specifying input folder
Currently, I would have to run the app multiple times as the scanner outputs files into separate directories for each date.  I want to have the option to either specific a  root path and add a flag to recurse.   If this option is enabled, it should also re-create the relative directory structure in the target output directory.

naglepuff:
Fix #170 

Adds a new flag (`-r` or `--recursive`).

If this flag is provided to the `imagedephi run` command, then redaction will be performed on sub directories of the input directory. Additionally, the directory structure will be duplicated in the output directory.

### DIFF
--- a/imagedephi/main.py
+++ b/imagedephi/main.py
@@ -112,12 +112,14 @@ def imagedephi(
     type=click.Path(exists=True, file_okay=False, readable=True, writable=True, path_type=Path),
 )
 @click.option("--rename/--skip-rename", default=True)
+@click.option("-r", "--recursive", is_flag=True, help="Redact images in subdirectories.")
 @click.pass_obj
 def run(
     obj: ImagedephiContext,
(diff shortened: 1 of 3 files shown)


manthey:
When you have --skip-rename turned on, it creates the directory structure (at least one deep) from the source, but the redacted files are not in those subdirectories.

naglepuff:
> When you have --skip-rename turned on, it creates the directory structure (at least one deep) from the source, but the redacted files are not in those subdirectories.

I forgot to update the parent directory of the output file when `--skip-rename` is used. I fixed that with 71119f8. 

I also updated the test to make sure this interaction doesn't break in the future with 231f38c.

manthey (approved):

PR #179, opened 2024-01-04
Use hatchling instead of setuptools

marySalvi:


brianhelba (approved):

tox.ini
```diff
@@ -8,10 +8,11 @@ env_list =
 
 [testenv]
 base_python = python3.11
-use_develop = true
+# Building and installing wheels is significantly faster
```
manthey, line 11: I thought the virtue of `use_develop` was that a change to the source didn't require rebuilding the tox environment to get the changes.  Does this always rebuild the wheels on source changes between test runs?

brianhelba: Yes, the distribution will be rebuilt from scratch each time the tests are run, which is more rigorous, since more of the packaging stack is being exercised.

However, with a combination of Tox v4 (using PEP 517 hooks), Hatchling, and wheel (instead of sdist) builds, the overhead of doing a full package build is now very small.

brianhelba: Just to be clear, the full Tox environment isn't being rebuilt. Dependencies installed once will be pre-installed for future runs. It's just the `imagedephi` wheel distribution which is being rebuilt each time, but that's now so fast that it doesn't matter.

manthey: Thanks.

PR #186, opened 2024-02-02
Update tests

marySalvi:
This PR updates the current tests to work properly with the new endpoint setup.
depends on #176

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -1,9 +1,9 @@
+import asyncio
 from collections.abc import Generator
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 import subprocess
 import sys
-from time import sleep
 
 from click.testing import CliRunner
 from freezegun import freeze_time
@@ -11,13 +11,7 @@ import httpx
 import pytest
 
 from imagedephi import main
-
-
-@pytest.fixture
-def thread_executor() -> Generator[ThreadPoolExecutor, None, None]:
-    executor = ThreadPoolExecutor(max_workers=1)
-    yield executor
-    # executor.shutdown(cancel_futures=True)
+from imagedephi.utils.network import wait_for_port
 
 
 @freeze_time("2023-05-12 12:12:53")
@@ -64,11 +58,16 @@ def test_e2e_gui(
 ) -> None:
 
     port = unused_tcp_port
-    subprocess.Popen(
-        [sys.executable, "-m", "imagedephi", "gui", "--port", str(port)],
 
+    gui = subprocess.Popen(
+        [sys.executable, "-m", "imagedephi", "gui", "--port", str(port)],
     )
-    sleep(2)
+
+    asyncio.run(asyncio.wait_for(wait_for_port(port), timeout=2))
+
+    # Check that the GUI is running
+    assert gui.poll() is None
+
     check_gui = httpx.get(f"http://127.0.0.1:{port}")
     assert check_gui.status_code == 200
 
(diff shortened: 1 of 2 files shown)


tests/test_e2e.py
```diff
@@ -58,56 +57,35 @@ def test_e2e_plan(cli_runner: CliRunner, data_dir: Path, rules_dir: Path) -> Non
...
-    # 500 error should cause the CLI to halt. However, if halting fails, then this test is
-    # particularly susceptible to exceeding the timeout, which on Windows will crash pytest (due
-    # to the behavior of pytest-timeout).
-
-    assert cli_result.exit_code == 0
-    webbrowser_open_mock.assert_called_once()
-    output_file = tmp_path / "Redacted_2023-05-12_12-12-53" / "study_slide_1.tif"
+
+    port = unused_tcp_port
+    subprocess.Popen(
```
brianhelba, line 67: What happens to this subprocess when this test case ends? Does it keep running in the background until the parent Pytest terminates?

tests/test_e2e.py
```diff
@@ -58,56 +57,35 @@ def test_e2e_plan(cli_runner: CliRunner, data_dir: Path, rules_dir: Path) -> Non
...
+        [sys.executable, "-m", "imagedephi", "gui", "--port", str(port)],
+
+    )
+    sleep(2)
```
brianhelba, line 71: So will this test always take 4 seconds at minimum, given the sleeps?

Is there a way to make it closer to the previous behavior, where it polled for an open socket or the termination of the CLI (or maybe just the existence of a output file)?

If test cases contain `sleep`, they're guaranteed to be much slower than every other test case for local development (causing overall test suite times to go up, sometimes by an order of magnitude). Sometimes in a CI environment with very limited resources, the hardcoded sleep duration is not enough, and the test will fail.

If there's no way around `sleep`, there's an option to use [custom Pytest markers)(https://docs.pytest.org/en/7.1.x/example/markers.html) to create groups like `slow` or `flaky`, which can be excluded for some test runs. Alternatively, you can use [OS-specific test cases](https://github.com/DigitalSlideArchive/ImageDePHI/blob/28dad54eb8fbf9f28db8873fd004baf91454a37e/tests/test_utils_os.py#L18) to include or exclude tests based on the OS, or even an OS-specific parameterized fixture which provides setup and teardown in different ways for a single consolidated test case.

tests/test_e2e.py
```diff
@@ -1,24 +1,23 @@
...
 import httpx
 import pytest
-from pytest_mock import MockerFixture
 
 from imagedephi import main
-from imagedephi.utils.network import wait_for_port
 
 
 @pytest.fixture
 def thread_executor() -> Generator[ThreadPoolExecutor, None, None]:
```
brianhelba, line 17: If you're not using a fixture, maybe it should be deleted, since otherwise it's mysterious cruft that everyone else will be afraid to delete too.

marySalvi: removed in c95f49ee9c4caa7ae50536b73b038fcb266dcc81

tests/test_e2e.py
marySalvi: fixed in c95f49ee9c4caa7ae50536b73b038fcb266dcc81

tests/test_e2e.py
marySalvi: explicitly killed in c95f49ee9c4caa7ae50536b73b038fcb266dcc81

manthey (approved):
This works well enough for now.

PR #187, opened 2024-02-09
Make recursive flag available for the `plan` subcommand

naglepuff:
Can be invoked with `imagedephi plan -r <input_directory>`

During testing of this change I noticed that the logger wasn't being correctly configured. It seems that the `fileConfig` function couldn't find the default `logging.conf` file at its location, so I moved it into the `/imagedephi` directory, and now it appears to be packaged correctly.

Additionally, invoking `run` without setting `-v` will increase log verbosity, since the plan outputs redaction steps at the `INFO` level and above.


imagedephi/main.py
```diff
@@ -38,6 +38,9 @@
         help="Path where log file will be created",
         type=click.Path(path_type=Path),
     ),
+    click.option(
+        "-r", "--recursive", is_flag=True, help="Apply the command to images in subdirectories"
```
manthey, line 42: `-r` is both `--recursive` and `--override-rules`

manthey (approved):

PR #202, opened 2024-03-15
Handle non-standard file extensions

marySalvi:
Defaults to tiff if bits match but file extension doesn't. Also reports downstream errors without stopping process


marySalvi:
```
def build_redaction_plan(
    image_path: Path, base_rules: Ruleset, override_rules: Ruleset | None = None
) -> RedactionPlan:
    file_format = get_file_format_from_path(image_path)
    if file_format == FileFormat.TIFF:
        # Since SVS is a subset of tiff, fall back on file extension
        file_extension = FILE_EXTENSION_MAP[image_path.suffix] if image_path.suffix in FILE_EXTENSION_MAP else file_format
        if file_extension == FileFormat.TIFF:
            merged_rules = base_rules.tiff.copy()
            if override_rules:
                merged_rules.metadata.update(override_rules.tiff.metadata)


            return TiffRedactionPlan(image_path, merged_rules)
        elif file_extension == FileFormat.SVS:
            merged_rules = base_rules.svs.copy()
            if override_rules:
                merged_rules.metadata.update(override_rules.svs.metadata)
                merged_rules.image_description.update(override_rules.svs.image_description)
            return SvsRedactionPlan(image_path, merged_rules)
        else:
            raise UnsupportedFileTypeError(f"File format for {image_path} not supported.")
    elif file_format == FileFormat.DICOM:
        dicom_rules = base_rules.dicom.copy()
        if override_rules:
            dicom_rules.metadata.update(override_rules.dicom.metadata)
            dicom_rules.delete_custom_metadata = override_rules.dicom.delete_custom_metadata
        return DicomRedactionPlan(image_path, dicom_rules)
    else:
        raise UnsupportedFileTypeError(f"File format for {image_path} not supported.")
```

I wonder if it would be better to log errors here instead of raising them because then are being handled upstream when this function is called and also when `show_redaction_plan` is called, which causing repeated code but also repeated logging when redacting.

What are your thoughts @manthey @naglepuff ?

imagedephi/redact/build_redaction_plan.py
```diff
@@ -35,7 +35,11 @@ def build_redaction_plan(
     file_format = get_file_format_from_path(image_path)
     if file_format == FileFormat.TIFF:
         # Since SVS is a subset of tiff, fall back on file extension
-        file_extension = FILE_EXTENSION_MAP[image_path.suffix]
+        file_extension = (
```
manthey, line 38: As a follow on issue, we should detect Aperio files based on the value of the ImageDescription tag and not the svs suffix.

manthey (approved):

PR #203, opened 2024-03-26
Image data table

marySalvi:
Shows redaction plan for images in the input directory when selected.
[Screencast from 04-04-2024 04:04:58 PM.webm](https://github.com/DigitalSlideArchive/ImageDePHI/assets/40494088/b0d87ff2-74db-41a6-a6fb-7152b55970d7)


manthey:
Each file has its own data table of what gets redacted or kept.  Not necessarily in this PR, but is there a way to have a unified table so that all the columns line up?

We eventually want to sort the entries in the table so that it is more obvious what is being redacted (i.e., I'd like to see the things being removed to the left of the things being kept).

manthey (approved):

PR #212, opened 2024-04-10
Yarn4 migration

marySalvi:
This moves the project from yarn version 1 to yarn version 4.5.1. It uses the "[plug 'n' play](https://yarnpkg.com/features/pnp)" installation workflow which eliminates the `node_modules` directory and uses a global cache for packages. 

**Setup**
- Must be using Node 18+
- run `corepack enable`
- ~~run `yarn set version berry`~~ -->This step shouldn't be necessary 
- run `yarn`

You may need to manually delete your `node_modules` directory.


client/.gitignore
```diff
@@ -22,3 +22,12 @@ dist-ssr
...
 *.sw?
+
+# Yarn directories
+node_modules/*
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/versions
```
marySalvi, line 33: Followed docs from https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

manthey:
We should add the `corepack enable` line to docs/development.md file.  Also, we probably should assert that we expect python 3.11, you need node >= 20, and yarn (do you need to specify yarn version berry to run tox -e binary or just tox?).

marySalvi:
> We should add the corepack enable line to docs/development.md file. Also, we probably should assert that we expect python 3.11, you need node >= 20

agreed

client/package.json
```diff
@@ -36,5 +36,11 @@
     "vite": "^4.5.5",
     "vue-eslint-parser": "^9.4.2",
     "vue-tsc": "^1.8.5"
+  },
+  "packageManager": "yarn@4.5.1+sha256.dc2e45a77f87a9c2a76708d098ab54ca0cfdd5ec2cf7d9e3c1f24bf786a2383c",
```
marySalvi, line 40: > (do you need to specify yarn version berry to run tox -e binary or just tox?).

Locally I am able to run `tox` and `tox -e binary` in a terminal with `yarn v1.22.0` and still have it build correctly so long as I am using `node 20+` with `corepack` enabled. I believe this line is why.

marySalvi: After some investigation, yes this line combined with `corepack` is the magic. Once it is enabled node is able to detect the correct yarn version. I have updated the ci yaml to reflect this.

Corepack is also what allows us to switch to different projects using different versions without manually setting a new version each time. Huzzah!

https://nodejs.org/api/corepack.html#configuring-a-package

manthey (approved):

naglepuff (approved):

PR #213, opened 2024-04-22
Delete Overview and Label DICOM Images

### LINKED ISSUE #195: Redact DICOM associated images

naglepuff:
Provide support for identifying and redacting (through deletion only) associated images that are part of DICOM slides. Additional redaction methods to follow in subsequent PRs.

Fix #195


naglepuff:
Previous attempts at running the tests on MacOS failed. Rebasing off of `main` appears to have fixed this problem.

manthey (approved):

manthey:
In theory, one could have a single dicom file will all the tile levels and the macro/label image.  If we find such a file, we will want to revisit how to redact those sub-images.

PR #218, opened 2024-05-07
Use tqdm for progress bar

### LINKED ISSUE #216: Provide Feedback During DeID CLI process
Would be useful to have some sort of TQDM type output at the folder and/or file level.  I was de-identifying about 170 images (400 GB) and it seems to be chugging along, but I have no simple way to figure out if it's working or crashed.  Using some linux tricks I can tell it's working ( du -ch  / etc..) but for a less technical user, they wouldn't have any way to figure out what's going on.

image

naglepuff:
The [tqdm](https://tqdm.github.io/) progress bar supports [redirection of logging](https://tqdm.github.io/docs/contrib.logging/#logging_redirect_tqdm) in order to keep the progress bar in place while other application output is printed for the user. This is especially helpful for running the CLI with the verbose (`-v`) flag.

Fix #216

### DIFF
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -3,10 +3,8 @@ from __future__ import annotations
 from collections.abc import Generator
 import datetime
 import importlib.resources
-from io import StringIO
 from pathlib import Path
 
-import click
 import tifftools
 import tifftools.constants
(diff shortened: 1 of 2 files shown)


marySalvi (approved):
LGTM!

PR #221, opened 2024-05-20
Add tiff and svs minimum tags

### LINKED ISSUE #165: Strict redaction - define minimum required meta
Define the minimum required metadata necessary to load .svs. This will allow users to appy a "strict" redaction profile without requiring the institution to inspect a bunch of tags. Start with .svs since the viewers and dataloaders are more mature.

naglepuff:
Fix #165 

This PR introduces a new "mode" for redaction alongside a new set of rules used for the new mode. By using the `--strict` flag with the `imagedephi run` command, users can now redact tiff-based whole slide images (.tiff, .svs) down to the minimum set of metadata so that the images can be read as tiff files.

## minimum_rules.yaml
If the user supplies the `--strict` flag, then a new ruleset called `minimum_rules.yaml` is used for redaction. This rule set contains tiff and svs sections with `keep`-type rules for all required metadata tags and a fallback rule that deletes all other tags. Associated images are also deleted.

## redact.py
The `redact_images` function now takes an additional keyword argument, `strict`. The function uses this argument to get the correct set of rules, and inform the creation of the redaction plan objects. For example, the `get_base_rules` function is passed this value and either returns a `Ruleset` object created from `base_rules.yaml` or `minimum_rules.yaml` based on the value of `strict`. Additionally, DICOM files are skipped in this mode. There is potential for future support in the mode for DICOM images when a minimum ruleset is determined.


manthey:
The strict flag should be available in the `plan` subcommand.

manthey (approved):

PR #222, opened 2024-05-21
Write a csv summary of redaction runs

### LINKED ISSUE #220: Emit file mapping input files to output files
Could be implemented as a CLI option to start.

naglepuff:
Fix #220 

Create a file that is the sibling of the redacted directory. The file is a csv and contains a mapping of input file names to output file names.

This change replaces the `create_redact_dir` function with `create_redact_dir_and_manifest`, which functions nearly identically except it also creates (via `<path>.touch()`) the manifest file and returns a 2-value tuple `(output_dir, manifest_path)` where both values are paths. The test written for `create_redact_dir` has been updated to reflect this change.

TODO:

- [x] Write tests

### DIFF
--- a/imagedephi/redact/redact.py
+++ b/imagedephi/redact/redact.py
@@ -67,7 +67,7 @@ def create_redact_dir(base_output_dir: Path, identifier: str | None = None) -> P
     """
     if not identifier:
         identifier = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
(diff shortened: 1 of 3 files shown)


naglepuff:
@marySalvi if you could PTAL and make sure these changes make sense for the UI workflow (I don't think there would be any impact but I'm not 100% sure)

@manthey if you could PTAL and let me know if you think that this simple manifest file is sufficient for now.

manthey (approved):

PR #223, opened 2024-06-04
Rename `match_counts` to be more clear

naglepuff:
@marySalvi from the discussion today it was clear that this variable name was unclear. I think changing it from `match_counts` -> `associated_image_count_by_rule` makes a big difference. Let me know if you have a different suggestion.


marySalvi (approved):
I think this is good!

PR #224, opened 2024-06-11
Get thumbnails for non-tiled images

### LINKED ISSUE #219: Support generating thumbnails for non-tiled images
If there are no tiled IFDs, we could try to use PIL to read the tiff.

naglepuff:
Fix #219 

#### Changes

Add a fallback for obtaining thumbnail images for tiffs. In the case that no lowest-resolution tiled IFD can be found, pass the image path to a new function , `get_image_response_from_tiff`. This new function attempts to use `PIL` to open up the entire image and generate a JPEG thumbnail from there.

Note that this does require disabling the `MAX_IMAGE_PIXELS` for `PIL.Image` in this function, since non-tiled tiffs can still be quite large, and `PIL` will throw an exception if the number of image pixels exceeds a certain threshold. I believe this is safe for now, since the intended use case is redacting images local to the user's machine.

#### To test

1. Run `imagedephi gui` to start the web server
2. Open your browser to `localhost:<port>`, where `<port>` is the port where the web application is running
3. navigate to `/?file_name=<path/to/file>&image_key=thumbnail` where `<path/to/file>` is the full path to a non-tiled tiff image
4. You should see a thumbnail-sized image in your browser

### DIFF
--- a/imagedephi/gui/utils/image.py
+++ b/imagedephi/gui/utils/image.py
@@ -98,7 +98,13 @@ def get_image_response_from_tiff(file_name: str):
     """
     Use as a fallback when we can't find the best IFD for a thumbnail imate.
     This happens when attempting to extract a thumbnail from a non-tiled tiff.
+
+    Since we completely trust the input (images on the current machine), we disable
+    max image size. This should probably be revisited in the future if ImageDePHI needs
+    to work over the web.
     """
+    max_size = Image.MAX_IMAGE_PIXELS
+    Image.MAX_IMAGE_PIXELS = None
     jpeg_buffer = BytesIO()
     image = Image.open(file_name)
     scale_factor = MAX_ASSOCIATED_IMAGE_HEIGHT / image.size[1]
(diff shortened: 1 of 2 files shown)


imagedephi/gui/utils/image.py
```diff
@@ -94,6 +96,27 @@ def get_image_response_from_ifd(ifd: "IFD", file_name: str):
             return StreamingResponse(jpeg_buffer, media_type="image/jpeg")
 
 
+def get_image_response_from_tiff(file_name: str):
+    """
+    Use as a fallback when we can't find the best IFD for a thumbnail imate.
```
manthey, line 101: s/imate/image/

manthey:
Not in this PR, but we may wish to reconsider just scaling images by height.  There are some images that have extreme aspect ratios, and this will end up with wide images (I could also see having the methods / endpoints specify a max width and a max height rather than having it encoded as a constant).

imagedephi/gui/utils/image.py
naglepuff: Fixed!

manthey (approved):

PR #225, opened 2024-06-11
UI image data table

marySalvi:


imagedephi/redact/redact.py
```diff
@@ -163,6 +166,50 @@ def redact_images(
...
+    excluded_tags = [
+        "StripOffsets",
+        "StripByteCounts",
+        "FreeOffsets",
+        "FreeByteCounts",
+        "TileOffsets",
+        "TileByteCounts",
+        "JPEGIFOffset",
+        "JPEGIFByteCount",
+    ]
```
manthey, line 197: We should exclude showing `BigTIFF`, which isn't a tag but a format detail.

manthey:
The color of the scroll bars
image
is nearly impossible for my eyes to resolve the page area from the background (if I rotate the color space I can see it).

imagedephi/redact/redact.py
marySalvi: done in 808131bfef0434c339f8feb03d57c83834784114

marySalvi:
> The color of the scroll bars image is nearly impossible for my eyes to resolve the page area from the background (if I rotate the color space I can see it).

I did my best to recreate what you were seeing and adjust accordingly in 808131bfef0434c339f8feb03d57c83834784114. Can you check to see if this helps?

imagedephi/redact/svs.py
```diff
@@ -187,11 +188,29 @@ def report_missing_rules(self, report=None) -> None:
...
+        self,
+    ) -> dict[
+        str,
+        dict[
+            str,
+            int
+            | str
+            | dict[str, str | int | float | bytes | list[int | float] | dict[str, str | int]],
+        ],
+    ]:
```
marySalvi, line 201: Writing this out in multiple places seems very awkward to me. I tried to do to make a generic type that would be reusable but it looks like what I wanted to do is only available in [3.12](https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695) so I don't know if that is advisable.
@manthey Do you have a recomendation?

manthey: Can't you just define a type once and reuse it?  That is, can't you do
```
PlanType = dict[
    str,
    dict[
        str,
        int | str | dict[str, str | int | float | bytes | list[int | float] | dict[str, str | int]],
    ],
]
```
and then use `-> PlanType` for this method?

manthey:
> > The color of the scroll bars image is nearly impossible for my eyes to resolve the page area from the background (if I rotate the color space I can see it).
> 
> I did my best to recreate what you were seeing and adjust accordingly in [808131b](https://github.com/DigitalSlideArchive/ImageDePHI/commit/808131bfef0434c339f8feb03d57c83834784114). Can you check to see if this helps?

It definitely helps.

imagedephi/redact/svs.py
marySalvi: updated in 488d7e61674124925ac40b04c55a1449719228eb

imagedephi/redact/svs.py
```diff
@@ -204,12 +209,25 @@ def report_plan(self) -> dict[str, dict[str | int, str | int]]:
...
                     operation = self.determine_redaction_operation(rule, image_description)
                     logger.info(f"SVS Image Description - {key_name}: {operation}")
-                    report[self.image_path.name][key_name] = operation
+                    report[self.image_path.name][key_name] = {"action": operation, "value": _data}
                 continue
             rule = self.metadata_redaction_steps[tag.value]
             operation = self.determine_redaction_operation(rule, ifd)
             logger.info(f"Tiff Tag {tag.value} - {rule.key_name}: {operation}")
-            report[self.image_path.name][rule.key_name] = operation
+            if ifd["tags"][tag.value]["datatype"] == 7:
```
manthey, line 217: Rather than `7`, this should be something like `tifftools.Datatype.UNDEFINED.value`

manthey (approved):
I've made some recommendations, but this could be merged and those addressed in a future PR.

marySalvi:
> A few other columns we shouldn't list: `associated_image_count`, `JPEGTables`, `NewSubfileType`, `Photometric`, `PlanarConfig`, `Predictor`.

Added in c10c83366dd15d4d13e3bf19ca735b2ca3d61b4b

imagedephi/redact/svs.py
marySalvi: Fixed in 02dac497951c541ca3e3ac007eafd1d3c6e84882

stubs/tifftools/tifftools.pyi
```diff
@@ -24,13 +24,15 @@ class IFD(TypedDict):
     bigtiff: bool
     tagcount: int
 
+Data = str | bytes | list[int | float]
```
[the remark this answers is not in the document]
marySalvi: Fixed in 02dac497951c541ca3e3ac007eafd1d3c6e84882

PR #228, opened 2024-06-26
Allow setting max height and width for assoicated images

### LINKED ISSUE #155: Thumbnail sizing
There is a constant that represents the upper limit of the height of an image returned from the `/image` endpoint, but nothing for the width. We should:

- Also include a maximum width for thumbnail/associated images
- Allow image dimensions to be specified in the URL

From #154

naglepuff:
This adds two new query params to the `/image` endpoint: `max_height` and `max_width`.

They are used to determine how to scale an associated image down so that it fits the requested size. The default value is 160 pixels, and is defined by the constanct `MAX_ASSOCIATED_IMAGE_SIZE` (previously `MAX_ASSOCIATED_IMAGE_HEIGHT`).

This also ensures that when picking an IFD for thumbnail generation, it picks an IFD that is larger than the requested thumbnail size if one exists.

Fix #155


imagedephi/utils/tiff.py
```diff
@@ -96,21 +96,35 @@ def get_associated_image_svs(image_path: Path, image_key: str) -> IFD | None:
...
+
+        if image_width and (not lowest_width_seen or image_width < lowest_width_seen):
+            # If the IFD's width is less than the minimum width we've seen
+            # so far, clock it as the lowest res and update the minimum
+            # seen width
+            # Additionally, check to make sure that the image width and height
+            # are both larger than the height/width of the thumbnail we are creating
+            # Only do this if we've found a candidate IFD already.
+            if lowest_res_ifd and (
+                image_width > thumbnail_width and image_height > thumbnail_height
```
manthey, line 121: These should be `>=`.

imagedephi/utils/tiff.py
```diff
@@ -96,21 +96,35 @@ def get_associated_image_svs(image_path: Path, image_key: str) -> IFD | None:
...
         if tifftools.Tag.TileWidth.value not in ifd["tags"]:
             continue
 
         image_width = int(ifd["tags"][tifftools.Tag.ImageWidth.value]["data"][0])
-        if image_width and (not min_width or image_width < min_width):
-            min_width = int(image_width)
-            lowest_res_ifd = ifd
+        image_height = int(ifd["tags"][tifftools.Tag.ImageHeight.value]["data"][0])
+
+        if image_width and (not lowest_width_seen or image_width < lowest_width_seen):
```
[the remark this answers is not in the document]
naglepuff: These should be addressed in 0ae8695

imagedephi/utils/tiff.py
naglepuff: Fixed in 0ae8695

manthey (approved):

PR #229, opened 2024-07-02
Clean up yaml files we commit to source control

### LINKED ISSUE #56: Lint yaml files we commit to the repository

naglepuff:
This PR does 2 closely related things:

1. It renames files with `.yml` to `.yaml`. This appears in yaml.org's [FAQ](https://yaml.org/faq.html) as the recommendation, which is as close to an official guideline that I could find. In any case, it doesn't really seem to matter, as long as we're consistent within our project.
2. ~~Add a `lint_yaml` step to our `tox` config which uses the python package [yamllint](https://github.com/adrienverge/yamllint). This was very easy to get set up and has very good documentation for configuration, with sensible defaults.~~
3. Modify our existing `lint` and `format` tests in `tox` to add a step to lint and format `.yaml` files we add to the repository. We are using a tool called [yamlfix](https://lyz-code.github.io/yamlfix/), which is also used by the [large image repository](https://github.com/girder/large_image). 

Fix #56


tox.ini
```diff
@@ -23,6 +24,13 @@ deps =
 commands =
     flake8 {posargs:.}
 
+[testenv:lint_yaml]
```
manthey, line 27: I'd rather see this as just another command in the `lint` environment rather than in its own environment (and, if we have a format step, it'd just be a another command in the `format` environment).

manthey:
Inspired by this PR, I decided to lint and format yaml in another code base (large_image).  Because I wanted to auto-format yaml, I ended up using `yamlfix` rather than `yamllint`.  yamlfix is less pedantic that yamllint, and I had to be more explicit about directories to process (see https://github.com/girder/large_image/blob/master/tox.ini#L181), but I consider this better than hand-adjusting the yaml to make the linter pass.  I also found it nice that I could add it to pre-commit (https://github.com/girder/large_image/blob/master/.pre-commit-config.yaml#L101-L107).

I did modify some of yamlfix's defaults (https://github.com/girder/large_image/blob/master/pyproject.toml#L76-L79) to avoid some changes (the only one I feel is important is the sequence style)

naglepuff:
I'll take a look at those large image changes when I get a chance to come back to this!

naglepuff:
> Inspired by this PR, I decided to lint and format yaml in another code base (large_image). Because I wanted to auto-format yaml, I ended up using `yamlfix` rather than `yamllint`. yamlfix is less pedantic that yamllint, and I had to be more explicit about directories to process (see [girder/large_image@`master`/tox.ini#L181](https://github.com/girder/large_image/blob/master/tox.ini?rgh-link-date=2024-07-10T20%3A18%3A58Z#L181)), but I consider this better than hand-adjusting the yaml to make the linter pass. I also found it nice that I could add it to pre-commit ([girder/large_image@`master`/.pre-commit-config.yaml#L101-L107](https://github.com/girder/large_image/blob/master/.pre-commit-config.yaml?rgh-link-date=2024-07-10T20%3A18%3A58Z#L101-L107)).
> 
> I did modify some of yamlfix's defaults ([girder/large_image@`master`/pyproject.toml#L76-L79](https://github.com/girder/large_image/blob/master/pyproject.toml?rgh-link-date=2024-07-10T20%3A18%3A58Z#L76-L79)) to avoid some changes (the only one I feel is important is the sequence style)

I more or less followed your example here to implement the same thing, with the same changes you made to the default configuration to keep the code bases in sync. One change I had to make here was with invoking `yamlfix` itself. Since we run our `tox` suite on windown, I couldn't get away with using `find` to selective pass relevant files to `yamlfix`. I had to use `yamlfix`'s `--exclude` option, which works but adds some additional overhead when run locally.

manthey (approved):

manthey:
I looked at how yamlfix does its exclude globbing and it is inefficient.  It calls pathlib.Path.glob on each of the exclude patterns for every yaml file found in the source tree.  This means if there are many candidate files that are excluded, this can become glacial.  I might make a PR to yamlfix to see if I can get it to adopt a less time consuming strategy.

PR #234, opened 2024-07-29
Redaction results table

marySalvi:
[Screencast from 07-29-2024 06:23:20 PM.webm](https://github.com/user-attachments/assets/39980f2c-de32-4be3-bacf-df38a4f162a8)


manthey:
I noticed an odd behavior in the progress meter:
progress
The machine happened to be under fairly heavy load aside from ImageDePHI.

manthey:
Aside from the progress bar oddity, this looks good.

From a feedback issue (not this PR), I wonder if we should leave the list visible behind the progress bar and update each row as redaction occurs (leaving it scrollable or auto-scrolling would be nice, but then we'd have to figure out how to have a blocking progress and scrollable make sense).

Also, as another future work, when we select a new directory to redact, we should probably hide the previous directory's listing so that it is is blank (possibly with progress of some sort) until the new directory shows up.  For long directories, this can be quite some time without any indication that _something_ is happening or that the current list is no longer relevant.

marySalvi:
> I noticed an odd behavior in the progress meter: progress [ ![progress]The machine happened to be under fairly heavy load aside from ImageDePHI.

fixed in 13a2b6b38b66860c05f880d0c11a26e400c5ed38

This was caused by the 'heartbeat' messages resetting the count value. In another PR I will make the ws ping messaging more sophisticated.

PR #235, opened 2024-07-30
Add redaction profile that modifies dates

### LINKED ISSUE #231: Add ability to change dates to January 1st
Currently our `base_rules.yaml`  rule set represents the most strict redaction options. Since the change proposed here would introduce a more lenient type of redaction (preserving years), it doesn't make sense to modify the base rule set at this time.

We could implement this at the tag level with a new redaction operation, say `retain_modified_dates`, and release a new rule set where every tag that represents a date has this redaction operation applied. Users could then extend this rule set however they see fit, either by copy/pasting and hand editing, or using the override rule set functionality. The bulk of the work here would be due diligence in making sure we capture every tag that is a date in the new rule set, and the actual algorithm that does this redaction (should it behave in a special way for certain fields like patient ages?)

There might be value in a setting at the image type or rule set level to use this redaction operation on all dates without specifying at each individual tag. We would then need to decide what to do if a date tag had a different operation (e.g. delete, keep) in that same rule set. There is some precedent in a higher-level rule. For DICOM images there is the `delete_custom_metadata` setting.

@manthey did you have an expectation of this based on how the current rule set definition works?

naglepuff:
Fix #231 

### Breaking Change
The `--strict` flag is no longer supported. It has been replaced by `--profile strict`.

### Changes
This PR continues work done to ship more default redaction profiles with Image DePHI. Specifically, a new base set of rules has been created to specially handle all known Date/Time attributes for tiff and DICOM files. These metadata tags are swapped out with semantically meaningful date (e.g. a tag holding a date will still hold a date post redaction), but are made "fuzzy" in the following ways:

- Dates: preserve the year, but set the month and day to January 1
- Times: set to midnight
- UTC Offsets: set to +0000

This redaction operation can be used in user-defined rule sets by specifying "modify_date" as the action for a given tag. If the value for that tag cannot be parsed as the expected type, it will be replaced by `None`.

Tests have been added for DICOM, Aperio, and tiff files. The tiff file used for testing had to be modified to include the `DateTime` tag, so the image has been reuploaded to our store, and the hash has been updated in `conftest.py`.


imagedephi/redact/tiff.py
```diff
@@ -197,10 +198,25 @@ def determine_redaction_operation(
...
 
+    def _get_modified_date(self, tiff_date: str) -> str | None:
+        """
+        Given a tiff datestring, return a version set to midnight, January 1st of that year.
+
+        Input should be  a string representing a date (formatted according to the tiff standard,
+        i.e. YYYY:MM:DD HH:MM:DD). If the given string does not conform to the given format,
+        return None.
+        """
+        expected_format = r"\d{4}:\d{2}:\d{2} \d{2}:\d{2}:\d{2}"
```
manthey, line 213: Do we want to add a `$` at the end of this match?  Otherwise, this allows arbitrary extra characters which are dropped (which might be fine -- I could be convinced either way, but we should say we are doing that if we intend it.)

naglepuff:
> We need to add tag 323 PlanarConfig to the tags kept in strict mode.

See #244

imagedephi/redact/tiff.py
naglepuff: See e4814ce. After thinking about it, I came to the conclusion that we should match the start of the string to the end, enforcing the expected pattern as strictly as we can.

naglepuff:
> Should we add a third enum to `--profile` (probably called `standard` or `default`)?

I added `"default"` as an option in 6d0e056. I think it adds some cohesion by giving the "no profile" mode a name. I made it the default value of `--profile` as well.

manthey (approved):

PR #239, opened 2024-08-01
Add flexibility for redacting custom DICOM tags

### LINKED ISSUE #188: Create third state for custom DICOM tag redaction
I think this ultimately should have three states, delete custom metadata, keep custom metadata, or do not have auto rules on custom metadata.  This could be a future PR, though.  Maybe it is a pair of booleans: delete_custom_metadata and keep_custom_metadata which would then make no sense if both true.

_Originally posted by @manthey in https://github.com/DigitalSlideArchive/ImageDePHI/pull/184#discussion_r1486722711_

naglepuff:
Fix #188 

### Breaking change
Any custom rule sets with `delete_custom_metadata` specified will need to be updated to use the new `custom_metadata_action` option.

### Changes
Replaces the `delete_custom_metadata` option in the rule set with `custom_metadata_action`. This can either be omitted, in which case a rule is needed for each piece of custom metadata in order for the image to be redacted, or set to `"keep"` or `"delete"`. If an action is specified, it will be applied if there is no rule for the custom metadata element. This is still only used for DICOM rule sets. The base rule set is configured to delete all custom metadata.

#### TODO 

- [x] Documentation
- [x] Tests


README.md
```diff
@@ -175,15 +175,19 @@ DICOM format rules are much the same:
...
     metadata:
         ...
 ```
 
-Note that here there is an eplicit format-level setting for dealing with custom metadata. Any tag with an odd group number is custom metadata. If not specified, this value default to `True`. Tag-level rules will override this behavior.
+Note that here there is an eplicit format-level setting for dealing with custom metadata. Any tag with an odd group number is considered custom metadata. This can be set to `keep`, `delete` or `use_rule`.
+
+* `keep`: Retain the custom metadata value after redaction.
+* `delete`: Delete the custom metadata tag from the image.
+* `use_rule`: This mode will fall back to rules specified for each piece of custom metadata in the `metadata` section of the rule set. If a custom metadata tag with no corresponding rule is encountered, the image will not be redacted, as the redaction plan would be considered incomplete.
```
manthey, line 187: Tag-level rules always apply even with `keep` or `delete`, right?  If so, we should mention that, too.

tox.ini
```diff
@@ -20,6 +20,8 @@ deps =
     flake8-docstrings
     flake8-isort
     pep8-naming
+    # https://github.com/lyz-code/yamlfix/issues/286
+    maison>=1.4.0,<1.4.3
```
manthey, line 24: yamlfix issued an update, so we can remove this now.

naglepuff: Removed in bd17bc3

manthey (approved):

PR #246, opened 2024-08-12
UI rules feature

### LINKED ISSUE #238: Add ability to add a new ruleset
Users need a way to include their own rulesets in the UI

marySalvi:
Closes #238 
Allows users to add and remove their own ruleset file and updates the redaction plan chart accordingly

### DIFF
--- a/client/src/HomePage.vue
+++ b/client/src/HomePage.vue
@@ -49,13 +49,15 @@ const redact_images = async () => {
   const response = await redactImages(
     selectedDirectories.value.inputDirectory,
     selectedDirectories.value.outputDirectory,
+    selectedDirectories.value.rulesetDirectory
   );
   if (response.status === 200) {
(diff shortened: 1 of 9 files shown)


manthey:
I used this to switch to the imagedephi/minimum_rules.yaml file, but my svs files still have entries like `LineCameraSkew` in them. I don't think this is directly a problem with this PR, as a command line run of `imagedephi -R imagedephi/minimum_rules.yaml run /src/folder -o /dest/folder` has the same results.  Adding `--profile strict` works from the command line.  I'll make an issue for this and test with a different set in a bit.

manthey (approved):

manthey:
When you switch rules, the new redaction plan is computed in the background.  If this takes a long time, there is no indication that this is happening.  As a subsequent PR, we need to add some sort of feedback so that the user knows that the currently displayed redaction is not current.

PR #249, opened 2024-08-20
Yamlfix doesn't work with a recent package

manthey:
Pin maison so yamlfix will still work.  See issue 286 on yamlfix for details.  The pinned packages should be unpinned when that issue is addressed.


manthey:
Yamlfix released a new version, so this is no longer needed.

Issue #252, opened 2024-08-29
Add Error Handling for Bad files


dgutman:
I have a collection of "girder breakers", and when it came to a file it didn't like, instead of catching the error, it just aborts..

 File "imagedephi/redact/redact.py", line 135, in redact_images
  File "imagedephi/redact/build_redaction_plan.py", line 50, in build_redaction_plan
  File "imagedephi/redact/svs.py", line 97, in __init__
imagedephi.redact.svs.MalformedAperioFileError

<img width="1470" alt="image" src="https://github.com/user-attachments/assets/053dab11-d74f-44c2-b03a-adcdf09c1d87">

manthey:
It looks like the proximate issue is that the attempt to log the problem file name and the error failed.  The underlying issue is that we used the file extension and this isn't a valid svs file (no Aperio ImageDescription); in this case, we should fall-back to just processing it as a tiff.

PR #253, opened 2024-08-29
Harden error logging

manthey:
Depending on the exception, the error logging itself could fail.


manthey:
This resolves the direct issue of #252.  Separately, we should also process tiff files that end in .svs as tiff files if they do not have an Aperio ImageDescription.

naglepuff (approved):

PR #254, opened 2024-09-03
Add `strict` to rule sets

naglepuff:
### Changes
Previously, "strict" redaction was done only by specifying a redaction profile from the command line. Now, rule sets themselves can be flagged as `strict`.

Effectively, this means that there is now no difference between running 

`imagedephi --profile strict ...` 
and 
`imagedephi -R minimum_rules.yaml ...`

Users can now create their own "strict" rule sets using this flag, and all tags not specified by that rule set would be deleted.

Documentation has been updated and a new test has been added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,14 @@ You can add a description to your custom rulesets. This is not used by the progr
 #### `output_file_name`
 Specify how the output files should be named here. The base ruleset contains the value `study_slide`. In this case, if the input slides are named: `john_smith_lung.svs` and `john_smith_pancreas.svs`, the redacted output images will be named `study_slide_1.svs` and `study_slide_2.svs`.
 
+### Other Top-level Properties
+
+#### `strict`
+The `strict` property of rulesets is used to denote that ALL unspecified tags should be deleted. This is supported for `tiff` and `svs` files. An example of using the strict flag can be seen in the `minimum_rules.yaml` rule set.
+
 ### File Format Rules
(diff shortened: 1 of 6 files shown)


manthey:
When I look at the plan for a set of svs files in `-R imagedephi/minimum_rules.yaml` versus using `--profile strict`, the `--profile strict` mode looks identical to the `--profile default` not to the minimum rules.  Specifically, the minimum rules say it will delete the ImageDescription and ImageDepth, whereas the strict and default both keep the ImageDescription tag (with some aperio metadata redacted).

naglepuff:
> When I look at the plan for a set of svs files in `-R imagedephi/minimum_rules.yaml` versus using `--profile strict`, the `--profile strict` mode looks identical to the `--profile default` not to the minimum rules. Specifically, the minimum rules say it will delete the ImageDescription and ImageDepth, whereas the strict and default both keep the ImageDescription tag (with some aperio metadata redacted).

I'm not able to reproduce this issue.

I created two files, on to store the output of `imagedephi -R imagedpehi/minimum_rules.yaml plan my_image.svs` and one for `imagedephi plan --profile strict my_image.svs`. After stripping off the timestamps from the log statements and running through a diff, I don't see any difference.

manthey:
I see the issue (I'll make a new issue report and approve this PR).

These two commands do NOT result in the same output:

`imagedephi --profile strict plan /my/directory`
`imagedephi plan --profile strict /my/directory`

If `--profile` comes before the command name, it is ignore because the default value after the command name takes precedence.

PR #258, opened 2024-09-11
Ping from frontend and reduce delay in messaging

### LINKED ISSUE #251: Websocket connection
After first redaction the websocket becomes unstable and doesn't automatically reconnect as expected

marySalvi:
Fixes #251 
Modelling after vite's [websocket](https://github.com/vitejs/vite/blob/587ad7b17beba50279eaf46b06c5bf5559c4f36e/packages/vite/src/client/client.ts#L175-L187) architecture, this implements a very simplified version:

- Periodically 'pinging' server
- Removes the server side 'heartbeat' 
- Reduces messaging delay (This is the most simplified portion since vite's HMR looks for changes in modules)

### DIFF
--- a/client/src/HomePage.vue
+++ b/client/src/HomePage.vue
@@ -38,6 +38,13 @@ ws.onmessage = (event) => {
     redact_dir: data.redact_dir || progress.value.redact_dir, // don't update if not present
   };
 };
+// Periodically ping the websocket
+
+setInterval(() => {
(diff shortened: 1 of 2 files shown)


manthey (approved):
While testing, I notice that the second set of redactions I do I briefly see the "Redacting images n/n" rather than the starting 1/n.  I think if we sent a 0/n message as soon as we start, this would resolve that.

PR #263, opened 2024-09-23
Redaction ux improvements

marySalvi:
This PR contains several smaller UI/UX improvements:

- Fixes sporadic white lines in the ImageData table header
- ImageData table remains visible during redaction
- Alters hover color on 'DE-PHI' button to not look disabled
- Resets progress counter to 0 on new redaction
- Redaction snackbar closes for new workflow (both new redaction and new input directory selection)
- Resets tags when new input directory is selected

### DIFF
--- a/client/src/HomePage.vue
+++ b/client/src/HomePage.vue
@@ -51,6 +51,8 @@ const redact_images = async () => {
     return;
   }
   redacting.value = true;
+  // Reset progress count
+  progress.value.count = 0;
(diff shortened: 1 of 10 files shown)


manthey (approved):

manthey:
One future change would be that when a new input directory is selected, we currently don't replace the table until it is parsed and there isn't any feedback that we are working on the new table.  Maybe the extant table should become greyed out or vanish or there should be some other indicator.

PR #277, opened 2024-11-21
Store selected directories in localStorage

marySalvi:
Saves the selected directories on update to localStorage and retrieves as the loading default.

### DIFF
--- a/client/src/HomePage.vue
+++ b/client/src/HomePage.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, onBeforeUnmount, onMounted } from "vue";
+import { ref, onMounted } from "vue";
 
 import { redactImages } from "./api/rest";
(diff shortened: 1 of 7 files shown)


marySalvi:
@manthey Would we want to clear these from storage at some point and if so under what circumstance?

manthey:
> @manthey Would we want to clear these from storage at some point and if so under what circumstance?

Provided we handle the directories no longer existing, and, in that case default to something sensible, I think it is okay to not delete them.

manthey:
We should show the images in the input directory if they are present.

manthey:
Probably nothing to do with this PR, but I notice that if I have the input directory dialog open and I click outside of it, it dismisses the dialog, leaves the navigated directory in the input directory field, but doesn't run the plan and show the contents of that directory.

manthey (approved):

PR #279, opened 2024-12-02
Create subcommand for getting demo data

naglepuff:
For now, just use the existing test image registry. A future commit will
switch to using some mouse data.


docs/demo.md
```diff
@@ -0,0 +1,183 @@
...
+In order to get the demo data, you will need to have installed Image DePHI and run the following command:
+
+```bash
+imagedephi demo-data
+```
+
+This will create a new directory in the location it is run called `demo_files` and download several whole slide images into that directory. These images contain fake PHI, which we will redact with Image DePHI.
+
+## Redacting with the Graphical User Interface (GUI)
+Image DePHI allows redaction of whole slide images through either a graphical user interface, accessible through a web browser, or a command line interface. First, let's take a look at the readaction workflow using the graphical user interface.
```
manthey, line 16: readaction -> redaction

docs/demo.md
```diff
@@ -0,0 +1,183 @@
+# Image DePHI Demo
+
+This walkthrough will guide you through using the Image DePHI program.
+
+## Getting the demo data
+
+In order to get the demo data, you will need to have installed Image DePHI and run the following command:
```
manthey, line 7: Elsewhere we use "ImageDePHI" as a single word.  I think I prefer that to having the space between Image and DePHI.

imagedephi/main.py
```diff
@@ -227,3 +232,36 @@ async def announce_ready() -> None:
...
+        reader = csv.DictReader(fd)
+        rows = [row for row in reader]
+        logger.info(f"Downloading files to {demo_file_dir}")
+        with logging_redirect_tqdm(loggers=[logger]):
+            for row in tqdm(rows, desc="Downloading demo images...", position=0, leave=True):
+                file_name = row["file_name"]
+                hash = row["hash"]
+                algo, hash_value = hash.split(":")
+                pooch.retrieve(
+                    url=f"http://data.kitware.com/api/v1/file/hashsum/{algo}/{hash_value}/download",
```
manthey, line 263: Change to https

PR #286, opened 2025-01-08
Lint pkg upgrades

marySalvi:
Updates Packages and fixes linting. 

Short summary of issues resolved:

- `vue-tsc` v1 appears to be broken and the recommendation is to upgrade to v2
- `eslint` v9 has [breaking changes](https://eslint.org/docs/latest/use/migrate-to-9.0.0) and the config file format completely changed. 
- It should be noted that `vue-tsc` does not work with the typescript  ^v5.7.2


manthey (approved):

manthey:
@marySalvi Thanks for sorting through this.

PR #290, opened 2025-02-25
288 yaml dephi input

### LINKED ISSUE #101: input_directory as variadac input argument
Update input directory to be a [variadic input argument](https://click.palletsprojects.com/en/8.1.x/arguments/#variadic-arguments).
This will also require updating the internal logic to perform based on various inputs:

- single file
- multiple files
- single directory
- multiple directories.

We should also have a discuss about how these different types of input can/should affect output.

### LINKED ISSUE #288: Bulk deidentification workflow improvements
We often have a single folder storing a large volume of slides and we want to deidentify only some of them. Under the current workflow we have to generate a copy of these files in another folder, and apply the deidentification there. This adds additional steps for the user (copy and deletion of temporary file), and the temporary volume of storage required can be on the order of hundreds of TB, and copying this can be quite slow. It is possible to script the calling of `imagedephi run` on individual files, but a lot of work is required to cleanup and aggregate the output.

Desired solution - consume a text file containing the select input files+paths to deidentify. This is also helpful in other scenarios (nested folders for example)

Allowing in place deidentification of the temporary copies is another solution, but this can lead to PHI leakage if some files fail.

marySalvi:
Closes #288 
Closes #101

- Updates `input_path` for variadic option input
- Adds ability for a user to pass a file with command details to `run` and `plan`
- Updates the demo to include this feature

### DIFF
--- a/docs/demo.md
+++ b/docs/demo.md
@@ -175,6 +175,36 @@ After that command finishes, you'll see a new directory in `./output_files` call
 
 You'll also see a file next to that directory called `Redacted_<timestamp>_manifest.csv`. This will contain a mapping of input file names to output file names, as well as any errors that may have occurred during redaction.
 
+### Using a command file in the CLI
+In some instances you may want to pass a command file to the CLI. For example you may have an long list of input files that would be cumbersome to type in a terminal.
+
+For this demo create a file called `command_file.yaml` and add the following:
+
+```bash
+---
+command: plan
+input_path:
+  - "demo_files"
+output_dir: ~/redacted_images
+```
+Now run the following:
+
+```bash
+imagedephi plan -c command_file.yaml
+```
+This option is supported by both the `plan` and `run` commands. Any option that can be added to these commands can also be added to the command file.
+
+```bash
+---
+command: run
+input_path:
+  - "demo_files"
+output_dir: /redacted_images
+recursive: True
+```
+**Please Note** The `command` line is not required but recommended to aid you in determining which options to use in the file.
+
+
 ## Next Steps
 
 For more information about the ImageDePHI rules system, be sure to check out the [documention](../README.md).
--- a/docs/demo.md
+++ b/docs/demo.md
@@ -202,9 +202,19 @@ input_paths:
 output_dir: /redacted_images
 recursive: True
 ```
-**Please Note** The command file is meant to supplement the command given in the terminal. Any option supplied in the terminal takes priority.
+ **Please Note:** The command file is meant to supplement the command given in the terminal. Any option supplied in the terminal takes priority.
 
-Additionally you can supply an unformatted yaml or text file. This will be treated as a list of  `input_files`.
+Additionally you can supply an unformatted yaml or text file with the `--file_list` option.
+
+Create a file called `file_list.txt` and add the following:
+```bash
+demo_files
+```
+Now run the following:
+
+```bash
+imagedephi plan -f file_list.txt
+```
 
 
 ## Next Steps
--- a/docs/demo.md
+++ b/docs/demo.md
@@ -197,12 +197,14 @@ This option is supported by both the `plan` and `run` commands. Any option that 
 ```bash
 ---
 command: run
-input_path:
+input_paths:
   - "demo_files"
 output_dir: /redacted_images
 recursive: True
 ```
-**Please Note** The `command` line is not required but recommended to aid you in determining which options to use in the file.
+**Please Note** The command file is meant to supplement the command given in the terminal. Any option supplied in the terminal takes priority.
+
+Additionally you can supply an unformatted yaml or text file. This will be treated as a list of  `input_files`.
 
 
 ## Next Steps
--- a/imagedephi/command_file.py
+++ b/imagedephi/command_file.py
@@ -3,5 +3,4 @@ from pydantic import BaseModel
 
 # The only two required fields are command and input_paths
 class CommandFile(BaseModel):
-    command: str
     input_paths: list[str]
--- a/imagedephi/command_file.py
+++ b/imagedephi/command_file.py
@@ -1,0 +1,7 @@
+from pydantic import BaseModel
+
+
+# The only two required fields are command and input_paths
+class CommandFile(BaseModel):
+    command: str
+    input_paths: list[str]
--- a/imagedephi/main.py
+++ b/imagedephi/main.py
@@ -204,6 +204,13 @@ def run(
 @imagedephi.command(no_args_is_help=True)
 @global_options
 @click.argument("input-path", type=click.Path(exists=True, readable=True, path_type=Path), nargs=-1)
+@click.option(
+    "-c",
+    "--command-file",
+    type=click.Path(exists=True, readable=True, path_type=Path),
+    help="File containing redaction command. "
+    "[INPUT_PATH] must be provided as an argument or in the command file.",
+)
 @click.pass_context
 def plan(
     ctx,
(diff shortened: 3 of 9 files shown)


manthey:
It would be nice to have a documented example and to make sure the input file can also be passed to the plan command.

manthey:
This doesn't quite work the way I expected.  I have a command file with two paths (e.g., 
```
input_path:
- /mnt/data3/ImageDePHI/SampleSet/TCGA-02-0321-01Z-00-DX1.3c57c27d-4a4f-45af-bf8b-2bde421a42fa.svs
- /mnt/data3/ImageDePHI/SampleSet/TCGA-06-0130-01Z-00-DX1.0391b65f-4e1d-4444-abb0-e5804606d461.svs
```
and I tried `imagedephi run -o /tmp/a -c test.yaml`, but it fails to use my output directory.  I think that the command file should supplement what is on the command line.

Further, for added convenience, I wonder if when you specify a command file that isn't valid yaml, if we should treat it as a simple line-separated list of input paths.

marySalvi:
> ```
> input_path:
> - /mnt/data3/ImageDePHI/SampleSet/TCGA-02-0321-01Z-00-DX1.3c57c27d-4a4f-45af-bf8b-2bde421a42fa.svs
> - /mnt/data3/ImageDePHI/SampleSet/TCGA-06-0130-01Z-00-DX1.0391b65f-4e1d-4444-abb0-e5804606d461.svs
> ```
> 
> and I tried `imagedephi run -o /tmp/a -c test.yaml`, but it fails to use my output directory. I think that the command file should supplement what is on the command line.

I am actually seeing the opposite locally. 
When I supply an output directory to the command line it is honored but if it is supplied through the command file it is not.
I forgot that we have a default for that option:
```
@click.option(
    "-o",
    "--output-dir",
    default=Path.cwd(),
    show_default="current working directory",
    help="Path where output directory will be created.",
    type=click.Path(exists=True, file_okay=False, readable=True, writable=True, path_type=Path),
)
``` 
I will have to remove the default, at least when a command file is present. 
Does `/tmp/a` exist? I'm not sure why its not being used otherwise.

manthey:
> > ```
> > input_path:
> > - /mnt/data3/ImageDePHI/SampleSet/TCGA-02-0321-01Z-00-DX1.3c57c27d-4a4f-45af-bf8b-2bde421a42fa.svs
> > - /mnt/data3/ImageDePHI/SampleSet/TCGA-06-0130-01Z-00-DX1.0391b65f-4e1d-4444-abb0-e5804606d461.svs
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and I tried `imagedephi run -o /tmp/a -c test.yaml`, but it fails to use my output directory. I think that the command file should supplement what is on the command line.
> 
> I am actually seeing the opposite locally. When I supply an output directory to the command line it is honored but if it is supplied through the command file it is not. I forgot that we have a default for that option:
> 
> ```
> @click.option(
>     "-o",
>     "--output-dir",
>     default=Path.cwd(),
>     show_default="current working directory",
>     help="Path where output directory will be created.",
>     type=click.Path(exists=True, file_okay=False, readable=True, writable=True, path_type=Path),
> )
> ```
> 
> I will have to remove the default, at least when a command file is present. Does `/tmp/a` exist? I'm not sure why its not being used otherwise.

The problem is that this line https://github.com/DigitalSlideArchive/ImageDePHI/blob/288-yaml_dephi_input/imagedephi/main.py#L180 fails if `output_dir` is not specified in the yaml file.  It should be guarded (e.g., something like `command_output = Path(command_params["output_dir"]) if "output_dir" in command_params else None`).

imagedephi/command_file.py
```diff
@@ -0,0 +1,7 @@
+from pydantic import BaseModel
+
+
+# The only two required fields are command and input_paths
+class CommandFile(BaseModel):
+    command: str
```
manthey, line 6: The docs say that `command` is not required, but this makes `command` required.

docs/demo.md
```diff
@@ -175,6 +175,36 @@ After that command finishes, you'll see a new directory in `./output_files` call
...
+
+```bash
+imagedephi plan -c command_file.yaml
+```
+This option is supported by both the `plan` and `run` commands. Any option that can be added to these commands can also be added to the command file.
+
+```bash
+---
+command: run
+input_path:
```
manthey, line 200: These should be `input_paths` to match the schema.

imagedephi/main.py
```diff
@@ -153,35 +167,110 @@ def run(
...
+                ]
+            if output_dir is None:
+                output_dir = Path.cwd()
+        else:
+            with command_file.open() as f:
+                command_params = yaml.safe_load(f)
+                try:
+                    CommandFile.model_validate(command_params)
+                except ValidationError:
+                    click.echo(
```
manthey, line 199: I wonder if we should support a different command line flag to always treat the file as a list (and then not get this warning).

imagedephi/command_file.py
marySalvi: Changed the docs here 96476b6a47508625dee80ce4126cfac4e1cde796

docs/demo.md
marySalvi: Changed here 96476b6a47508625dee80ce4126cfac4e1cde796

imagedephi/main.py
marySalvi: In this scenario would we want to still prompt a warning like this if that command line flag isn't supplied?

manthey: My though was that if we had two flags then the current `-c` would just print an error and stop and the new flag (maybe `-f` and `--filelist`) would expect just a list.

marySalvi: added in 17e09633d7a9e7d65761933198c444070a101731

imagedephi/main.py
```diff
@@ -153,35 +173,124 @@ def run(
...
+    cf_override_rules: Path | None = None
+    cf_profile: str = ""
+    # cf_rename will only be checked if rename is false so default should be false.
+    cf_rename: bool = False
+    cf_recursive: bool = False
+    cf_index: int = 1
+    if command_file:
+        if command_file.suffix not in [".yaml", ".yml"]:
+            click.echo(
+                "Command file does not match the expected format. \n"
```
manthey, line 195: If we print the same message in four locations, we should define it somewhere and reuse it.  Also, as long as we are using click.echo, we shouldn't do line breaks ourselves, just do `click.echo(click.text_wrap(<text>))`.

imagedephi/command_file.py
manthey: Hmm... since you might pass the a file list to plan and then to run, do we want to require `command`?

marySalvi: removed in c070497166748511c94699d2af53c5dc17f850e9

imagedephi/main.py
marySalvi: refactored in ce9d1869969f838b9d2adb48e249abbec0556c1a

imagedephi/main.py
```diff
@@ -24,6 +27,11 @@
 
 shutdown_event = asyncio.Event()
 
+fomat_message = (
```
manthey, line 30: fomat -> format

marySalvi: fixed  in b6e8bb9f84efa681723dffea0853dce933bf1fda

manthey (approved):

PR #335, opened 2026-07-20
Export associated images

cooperlab:
This adds the option to export associated images (label, macro, and thumbnail) to the cli and api. Core facilities that offer deidentification services may require the provenance provided by the label images, and may need to review thumbnails to confirm no PHI leakage into the image pyramid. The thumbnails are also helpful to the receiver of the data to organize and sort the deidentified images.

Refactoring was done to the utilities so that associated images can be accessed from within redact.py and to eliminate other circular imports.

To discuss:
The associated images are currently deposited in an "associated" subfolder within the output folder. There is risk of PHI disclosure if these are transferred to the recipient. May need a different location, or to more clearly label which folders do/not presumably contain PHI.


cooperlab:
Typing error are being raised for the IFD argument below. I'm not clear how the original code was handling this so I was hesitant to try to fix. Everything else is passing.

`imagedephi/redact/redact.py:156: error: Argument 1 to "get_image_bytes_from_ifd" has incompatible type "IFD | None"; expected "IFD"  [arg-type]
imagedephi/redact/redact.py:172: error: Argument 1 to "get_image_bytes_from_ifd" has incompatible type "IFD | None"; expected "IFD"  [arg-type]`

manthey:
This works fine in my testing.  I'll add a test in a follow on PR.

manthey (approved):

PR #349, opened 2026-08-13
New UI for associated images

### LINKED ISSUE #336: Expose rename, export associated options in UI
In some workflows, the person operating the UI may receive renamed files and want to preserve the filenames. They may also be required to review thumbnail images for PHI contamination, or to save file labels for provenance reasons.

PR #335 adds the ability to export associated images, and modifies the API redact function to expose this and the rename option. These could be made accessible within the UI by adding a "step 4 - export options" with toggle/checkboxes for file renaming and associated image export. It appears very simple to capture these element values and to pass them along to the redact API call.

manthey:
Closes #336

image

### DIFF
--- a/client/src/HomePage.vue
+++ b/client/src/HomePage.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, onMounted } from "vue";
+import { ref, onMounted, watch } from "vue";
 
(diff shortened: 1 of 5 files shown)


manthey:
@cooperlab I'll merge this once CI passes.  If you think there should be changes, let me know.

